PR #9, opened 2020-03-09
Add category recommandations test

acbeni:
Adding tests for category recommandations.


acbeni:
Yep, the idea here is just to validate that the URLs are correct, nothing deep at least at this stage. since i am more interested in completing coverage of the endpoints for this first iteration.

PR #16, opened 2020-03-16
chore: add missing data

acbeni:
Tip for review, skip the generated code and go straight for the tests 😄

### DIFF
--- a/packages/ml-sdk/src/generated/models/missing-data.ts
+++ b/packages/ml-sdk/src/generated/models/missing-data.ts
@@ -11,7 +11,11 @@
  *
  */
 
-import { ProductReference, ProductTypeReference } from './common'
+import {
+  ProductReference,
+  ProductTypeReference,
+  TaskStatusEnum,
+} from './common'
 
 export interface AttributeCount {
   /**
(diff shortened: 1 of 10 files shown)


packages/ml-sdk/src/generated/models/missing-data.ts
```diff
@@ -84,8 +74,8 @@ export interface MissingAttributes {
   readonly attributeCoverage?: AttributeCoverage
 }
 export interface MissingAttributesMeta {
-  readonly productLevel: MissingAttributesProductLevel
-  readonly variantLevel: MissingAttributesVariantLevel
+  readonly productLevel: MissingAttributesDetails
+  readonly variantLevel: MissingAttributesDetails
```
evilaz: why is this renamed? is it just internal? 🤔

acbeni: actually here the types are the same, they just have different descriptions, this is why i decided to make the same type for both.

acbeni:
Do you have any other questions here @evilaz ?

mbarga:
In addition to attributes we also have missing images and prices. Wonder if its also worth making tests for the other two endpoints?

acbeni:
I will add them in a separate PR @mbarga .

evilaz:
> Do you have any other questions here @evilaz ?

No I'm good, I don't get much anyway was just curious :)

PR #17, opened 2020-03-17
Complete missing data endpoint

acbeni:
This the last addition to missing data, missing prices and missing images


evilaz:
hm can't really review the logic and correctness. One question though, we need separate `MissingPricesTaskStatus` for each one? I mean because basically `TaskStatus` concept is shared across all Missing Data endpoints

acbeni:
@evilaz actually you are right, but for strongly typed languages, this helps since you can know what results you get.

acbeni:
i was leading the ts sdk at the client team too 😄 , but my idea is that no need to complicate the review for now, i ll just make a quick pass to aim for full coverage. next we ll try to enhance test coverages...

PR #35, opened 2020-11-04
fix use platform-sdk package instead of typescript-sdk

MicheleRezk:
fix tests to use platform-sdk package instead of typescript-sdk


packages/ml-sdk/package.json
```diff
@@ -25,7 +25,8 @@
     "@commercetools/sdk-client": "^2.1.1",
     "@commercetools/sdk-middleware-auth": "^6.0.4",
     "@commercetools/sdk-middleware-http": "^6.0.4",
-    "@commercetools/sdk-middleware-logger": "^2.1.1"
+    "@commercetools/sdk-middleware-logger": "^2.1.1",
+    "@commercetools/platform-sdk": "^1.4.0"
```
jenschude, line 29: shouldn't this be a deve dependency as it's only used in the tests?

PR #38, opened 2020-11-10
add baseUri in createApiBuilderFromCtpClient

MicheleRezk:
Add baseUri in createApiBuilderFromCtpClient so it can be configured while creating ApiRoot


packages/importapi-sdk/src/ctp/ctp-client.ts
```diff
@@ -1,6 +1,9 @@
 import { ApiRoot } from './../generated/index'
 
-export function createApiBuilderFromCtpClient(ctpClient: any): ApiRoot {
+export function createApiBuilderFromCtpClient(
+  ctpClient: any,
+  baseUri?: string
+): ApiRoot {
   return new ApiRoot({
     executeRequest: ctpClient.execute,
```
jenschude, line 8: baseUri is not forwarded to the ApiRoot

jenschude (changes requested):

jenschude (approved):

PR #123, opened 2021-09-27
Fix issues with SDK not working with Browsers

ajimae:
# Tasks

- [x] Add webpack and webpack configuration to bundle required packages to umd modules
- [x] Install required webpack plugins to facilitate umd bundling


jenschude (approved):
LGTM 👍

PR #147, opened 2021-11-23
chore(automation): add stale.yml with config

jherey:
This PR adds the stale bot to automatically close an issue after 7 days (still to be agreed by the team) of inactivity.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,16 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 5
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 2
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,16 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 5
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 2
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - blocked
```
ajimae, line 7: I think we can have a better name for the label. A name like `active` in my opinion should be more appropriate.

ajimae (approved):

Issue #217, opened 2022-03-14
No typescript sdk method for batching/pagination


lojzatran:
If we want to fetch all the resources, currently there is no way to do so with typescript sdk since the limit per query is 500 resources. There is [a way to use old nodejs client](https://github.com/commercetools/commercetools-sdk-typescript/blob/026d91a747b445b0913eabb7eb91ecc7f3eb257e/packages/sdk-client/test/client.test/sdk-client.test.ts#L209), but I would expect a similar method for typescript SDK as well. 

I would expect to have an API similar to this:
```js
await apiRoot
  .withProjectKey({ projectKey: '{projectKey}' })
  .customers()
  .get()
  .process((response) => doSth(response))
```
Even better there could be a method that would return a generator so that it could be used in loop:
```js
for await (const customers of apiRoot
  .withProjectKey({ projectKey: '{projectKey}' })
  .customers()
  .get()
  .fetchPages()
) {
  doSth(customers)
}
```
Something similar to what has already been done here: https://github.com/commercetools/commercetools-subscriptions/blob/a3a31d238d7e87cd789e3862efbb89a45463001d/src/create-subscription-orders.js#L48

PR #220, opened 2022-03-16
Expose Client Process Function

ajimae:
### Summary

- refactor the client to expose the process function at the client level
- write tests to ensure new and existing features still works properly

### Releated Issue
#217

### DIFF
--- a/packages/sdk-client/src/index.ts
+++ b/packages/sdk-client/src/index.ts
@@ -1,5 +1,5 @@
 export { default as ClientBuilder } from './client-builder/ClientBuilder'
-export { default as createClient } from './sdk-client/client'
+export { default as createClient, process as Process } from './sdk-client/client'
 export { default as getErrorByCode } from './sdk-client/errors'
 export { default as createAuthForAnonymousSessionFlow } from './sdk-middleware-auth/anonymous-session-flow'
 export { default as createAuthForClientCredentialsFlow } from './sdk-middleware-auth/client-credentials-flow'
--- a/packages/sdk-client/test/client.test/sdk-client.test.ts
+++ b/packages/sdk-client/test/client.test/sdk-client.test.ts
@@ -1,10 +1,12 @@
 import qs from 'querystring'
-import { createClient } from '../../src/sdk-client'
+import { Process, createClient, ClientBuilder } from '../../src'
+import { createApiBuilderFromCtpClient } from '../../../platform-sdk'
 import {
   ClientRequest,
   HttpErrorType,
   MiddlewareResponse,
 } from '../../src/types/sdk.d'
+import fetch from 'node-fetch'
 
 const createPayloadResult = (tot, startingId = 0) => ({
   count: tot,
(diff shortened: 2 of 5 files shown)


packages/sdk-client/src/index.ts
```diff
@@ -1,5 +1,5 @@
 export { default as ClientBuilder } from './client-builder/ClientBuilder'
-export { default as createClient } from './sdk-client/client'
+export { default as createClient, process as Process } from './sdk-client/client'
```
jherey, line 2: Any reason why it's been exported from the package with a capital "Process". I'm guessing the `process` function can be chained

ajimae: Had to make it `Process` with a capital `P` so it doesn't get shadowed by the node `process` object.

Sorry I don't understand what you mean by the `process` function can be chained, please elaborate more on this.

Thanks.

jherey (approved):

packages/sdk-client/test/client.test/sdk-client.test.ts
```diff
@@ -523,3 +525,372 @@ describe('process', () => {
...
+  describe('validate arguments', () => {
+    const middlewares = [
+      (next) =>
+        (...args) =>
+          next(...args),
+    ]
+
+    createClient({ middlewares }) as any
+    test('should throw if second argument missing', () => {
+      expect(() => Process(request, null, {})).toThrow(
```
lojzatran, line 547: Why `Process` is a separate function and not like this?
```
await apiRoot
  .withProjectKey({ projectKey: '{projectKey}' })
  .customers()
  .get()
  .process((response) => doSth(response))
```
I saw Jen's comment, but I do not understand what is the problem if it is like this? IMO it's more intuitive than to have a special function `Process`. Thanks.

jenschude: Cause process function embedded can be misleading. Not every request may be suitable to be processed as they can be inefficient.

jenschude: And process only makes sense for queries and not like every method at every endpoint across all APIs

lojzatran: Thanks, it makes sense, but how a static `Process` method will help in this case? Do you plan it to be like a 'hidden' feature that only if a user knows about it, they will use it?

> And process only makes sense for queries and not like every method at every endpoint across all APIs
Could TS help in this case so that `process` would be available only for queries?

jenschude: I think we will have to add some kind of interface in the future in order to limit it correctly. As like as we did for the java SDK.

It should not be a hidden feature, but it must be documented and promoted. The exposed process function should make it easier to use it without the need to have a client instance explicitly.

lojzatran: From a user's perspective it would be much easier to have an embedded function. We had it in the [old SDK](https://github.com/commercetools/sphere-node-sdk/blob/096ae225bf501103d679cda034b20d92ab31036f/src/coffee/services/base.coffee#L366) and it was working fine. 
But if you think there is a problem with the approach, what you have now is also ok for me.

ajimae: The only way to do it like [this](https://github.com/commercetools/sphere-node-sdk/blob/096ae225bf501103d679cda034b20d92ab31036f/src/coffee/services/base.coffee#L366) with our current setup is to add `a process` method in the `ApiRoot` class, which as Jens said is not the best thing to do.

For now, there are just two options that I can think of:

- Create a `client` and call the `process` method on the client just as we mentioned before
- Use this embedded `Process` function.

One other option would be to dynamically add the process method to the ApiRoot at runtime, but again we running from polluting the class and also this might introduce some unexpected behaviour.

lojzatran (approved):

packages/sdk-client/test/client.test/sdk-client.test.ts
jenschude: The linked old SDK just covered the platform API and introduced a Service concept for each domain. The newer SDKs are separating the client and the requests. This gives the opportunity to reuse the client configuration, as well as the requests itself and add middlewares much simpler. Also it's nearer to the nature of an HTTP api.

The request builders (api roots) are more of an helper to avoid memorizing all endpoints, parameters etc and to code exploratively. And it avoids breaking changes if new parameters are getting introduces or some names may change as the structure is not changing

PR #252, opened 2022-04-27
Better error message on failed http request

sfernengel:
When we get a 404, the returned error is empty and not helpful. Make sure we report at least the url and the status text in that case.

### DIFF
--- a/packages/sdk-client/src/sdk-middleware-auth/base-auth-flow.ts
+++ b/packages/sdk-client/src/sdk-middleware-auth/base-auth-flow.ts
@@ -92,12 +92,18 @@ async function executeRequest({
     }
 
     // Handle error response
-    let parsed
-    const text: any = await _res.text()
-    try {
-      parsed = JSON.parse(text)
-    } catch (error) {
-      /* noop */
+    let parsed: any;
+    let text: string;
+    if (!_res.ok) {
+      text = `${url}: ${_res.statusText}`;
+    } else {
+      text = await _res.text();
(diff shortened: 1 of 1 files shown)


packages/sdk-client/src/sdk-middleware-auth/base-auth-flow.ts
```diff
@@ -92,12 +92,18 @@ async function executeRequest({
...
     // Handle error response
-    let parsed
-    const text: any = await _res.text()
-    try {
-      parsed = JSON.parse(text)
-    } catch (error) {
-      /* noop */
+    let parsed: any;
+    let text: string;
+    if (!_res.ok) {
```
ajimae, line 97: We already made a check for `_res.ok` on line 59, I don't think this check here is necessary.

ajimae (commented):
I would like to know the response you got in your console when `.withLoggerMiddleware()` is included when creating the client.

sfernengel:
> I would like to know the response you got in your console when `.withLoggerMiddleware()` is included when creating the client.

I still get no error message:
node index.js
```js
Error
    at executeRequest (/Users/sevtap/Code/buggy_store/node_modules/@commercetools/sdk-client-v2/dist/commercetools-sdk-client-v2.cjs.dev.js:360:19)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
```

ajimae:
Changes no longer needed.

PR #278, opened 2022-06-20
Add Best Practices Markdown File to Repository

ajimae:
### Summary
- [x] add `BEST_PRACTICES.md` file to repository

### DIFF
--- a/BEST_PRACTICES.md
+++ b/BEST_PRACTICES.md
@@ -1,0 +1,387 @@
+# TypeScript SDK Best Practices
+
+## Contents
+
+- [General](#general)
+
+  - [Use up-to-date SDK](#use-up-to-date-sdk)
+  - [Documentation](#documentation)
+  - [Issues](#issues)
+  - [Resporting Issues](#reporting-issues)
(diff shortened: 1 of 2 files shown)


jherey (approved):

briantomkins:
Added a docs issue for the generic docs as well.
https://github.com/commercetools/commercetools-docs/issues/4750

briantomkins:
We recommend a timeout of 2 seconds for most simple requests.

However, for POST updates to carts, orders, and products times can vary depending on the use case, so we recommend using a large value that reflects the slowest performance of these calls in production plus some additional overhead.

For example, a project might have a 5 seconds timeout on payments extension then a typical p99 latency including transfer might be 6 seconds. Setting a 12 seconds timeout would be recommended to make these requests robust during times of slower end-to-end performance.

The maximum timeout supported is 60 seconds, also note that complex queries to fetch multiple resources will timeout in 20 seconds. Queries should be designed to typically complete in 5 seconds in average and return no more than 20mb of data in average. Please raise a support ticket if you are unable to achieve this.

PR #318, opened 2022-09-08
Support node16

barbara79:


.circleci/config.yml
```diff
@@ -16,14 +16,21 @@ aliases:
...
 executors:
-  node_14:
+  docker:
     docker:
-      - image: cimg/node:14.19.3
+      - image: cimg/node:14.20.0
 
 version: 2.1
+orbs:
+  node: circleci/node@5.0.2
```
ajimae, line 25: Do we really need this?

barbara79: yes, otherwise the build will fail I found this https://circleci.com/developer/orbs/orb/circleci/node

barbara79: I tried without but the built for each version of node was failing

ajimae: Well since you are using circleci orbs, this can pass.

Issue #319, opened 2022-09-12
withExistingTokenFlow should have second argument as optional


JohnMarsden24:
If you use the `withExistingTokenFlow` middleware TS forces you to provide it two arguments, the first being the `authorization` (token) and the second being `options` which only has one optional value of `force`:

```typescript
interface ExistingTokenMiddlewareOptions = {
  force?: boolean
}

withExistingTokenFlow(authorization: string, options: ExistingTokenMiddlewareOptions): ClientBuilder;
```

Which leads to my scenario when I had to to call the middleware like this:

```typescript
withExistingTokenFlow('random-token', {})
```

Happy to open the PR myself and contribute!

ajimae:
See #320

PR #320, opened 2022-09-12
fix: made options to withExistingTokenFlow optional

JohnMarsden24:
### What does this PR do?

Changes the `withExistingTokenFlow` type so that the second argument, `options` is optional.

### Why is this PR needed?

Currently as the second argument is required but you don't need to alter the options you need to pass an empty object to appease TypeScript:

`.withExistingTokenFlow('random-token', {})` 

[Mentioned in issue 319](https://github.com/commercetools/commercetools-sdk-typescript/issues/319)

### How was this PR achieved?

By making the second argument optional:

```typescript
 withExistingTokenFlow(
    authorization: string,
    options?: ExistingTokenMiddlewareOptions
  )
```

### DIFF
--- a/.changeset/violet-timers-sin.md
+++ b/.changeset/violet-timers-sin.md
@@ -1,0 +1,5 @@
+---
+"@commercetools/sdk-client-v2": patch
+---
+
+fix: make options for `withExistingTokenFlow` method optional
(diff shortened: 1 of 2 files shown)


ajimae (approved):
LGTM!

PR #333, opened 2022-09-21
chore(security): fixed security alert

barbara79:


examples/me/client/package.json
```diff
@@ -66,7 +66,7 @@
     "html-webpack-plugin": "^4.3.0",
     "webpack": "^5.53.0",
     "webpack-cli": "^4.8.0",
-    "webpack-dev-server": "^3.11.0",
+    "webpack-dev-server": "^4.11.1",
```
jherey, line 69: Is this a breaking change in the webpack dev server?

examples/package.json
```diff
@@ -8,6 +8,6 @@
   },
   "devDependencies": {
     "jest": "^27.2.5",
-    "puppeteer": "^10.4.0"
+    "puppeteer": "^18.0.4"
```
jherey, line 11: Same here. I don't know if the example still works, maybe you can test it?

examples/vue/package.json
```diff
@@ -14,8 +14,8 @@
     "vue": "^3.0.0"
   },
   "devDependencies": {
-    "@vue/cli-plugin-babel": "~4.5.0",
-    "@vue/cli-service": "~4.5.0",
+    "@vue/cli-plugin-babel": "~5.0.8",
+    "@vue/cli-service": "~5.0.8",
```
jherey, line 18: Also here, similar to a major version upgrade with vue.

jherey (commented):
Well done 👏🏽
Just some comments about the upgrade.

examples/me/client/package.json
barbara79: yes, but it is isolated just to the examples folder I think it wouldn't be a high impact on our SDK. WDYT?

examples/package.json
barbara79: the same answer here WDYT?

examples/react/package.json
```diff
@@ -30,7 +30,7 @@
     "html-webpack-plugin": "^4.3.0",
     "webpack": "^5.53.0",
     "webpack-cli": "^4.8.0",
-    "webpack-dev-server": "^3.11.0",
+    "webpack-dev-server": "^4.11.1",
```
barbara79, line 33: the same here too @jherey

Issue #353, opened 2022-10-21
Support persisting anonymous sessions between page reloads.


YGKtech:
Apologies if this isn't the right place for this, or if there's already a solution to my problem which I've somehow missed.

I'm trying to find a way to keep an anonymous user session alive after a user refreshes the page, e.g. so an anonymous user could load the page, add an item to a cart, reload/navigate away, then return and check out their cart.

I'm not seeing any mention of this topic in the docs, or any obvious support for it in the codebase. I did note that the anonymous auth flow accepts an anonymous_id parameter, so I tried capturing the id generated during the oauth flow, storing it in localStorage, and passing it in after a page reload, but this results in an error because the id is 'already in use'. I also tried passing in a custom tokenCache object which would read & write localStorage, but the SDK code never actually makes calls to my tokenCache (seems there's an old issue about this here: https://github.com/commercetools/commercetools-sdk-typescript/issues/140 which nobody ever responded to so it was auto-closed).
I imagine there may be a way for me to accomplish this by capturing the access token and passing it in to the ExistingToken auth flow, I'll likely try that next, though even if it works I'm not sure I'd consider it a production-ready solution, I'm having to jump through hoops to read the access token, and having to conditionally use a different auth flow is a bit clunky.

Normally when integrating an API/SDK like this, I would expect to implement this sort of functionality myself. So I'm really not opposed to having to do some of this work myself, but the SDK is already trying to automate a lot of the token handling, so it seems like a feature that should already be built in, and it doesn't expose the token value anywhere, so I had to pass in a modified copy of `fetch` to capture the token data.

ajimae:
Hi @YGKtech 

Thanks for bringing this up. For anonymous authentication flow, the anonymous ID is used to get an anonymous token which is then cached in the `tokenCache` using a `tokenCacheKey` [here](https://github.com/commercetools/commercetools-sdk-typescript/blob/70db08858547cfddc7601269a9a488cb85e5699d/packages/sdk-client/src/sdk-middleware-auth/base-auth-flow.ts#L69), all subsequent requests should now use the cached token instead provided it's still within the same request-response session. To a good extent SDK automates this process of token handling.

Another thing I would suggest is to capture and save the `anonymous token` gotten after the first request and then use it in the `headers` of the subsequent request, note that here you are now managing your own `tokenCache`.

For the issue about the SDK not making calls to the auth token will be investigated and see if we can reproduce it on our end and why that is the case.


Thanks.

[...]

PR #355, opened 2022-10-25
Fix TokenCache Store

ajimae:
### Summary
- [x] allow anonymous auth flow to accept custom tokenCache store
- [x] fix refresh token flow and password flow to accept custom tokenCache store also

### Related Issues
#353


jenschude (approved):
👍

Issue #354, opened 2022-10-21
Retry not fired on request timeout


olup:
It seems the client is not retrying when we are adding a timeout (using native node 18 `fetch`and `abortController`) - at least our datadog traces don't show any retries.

We get the "Network error: User aborted a request" error from CT client though.

If I understand correctly the code, it should retry independently of the `errorCodes` array because aborting will throw at fetch ("network") level, not as a http error, am I right ?

This is roughly the client and http middleware declaration, can you spot anything wrong with it ?

```
      const authMiddlewareOptions: AuthMiddlewareOptions = {
        host: `https://auth.${REGION}.gcp.commercetools.com`,
        projectKey: config.environment.COMMERCETOOLS_PROJECT_KEY,
        credentials: {
          clientId: config.environment.COMMERCETOOLS_CLIENT_ID,
          clientSecret: config.environment.COMMERCETOOLS_CLIENT_SECRET,
        },
        scopes: SCOPES,
      };


const httpMiddlewareOptions: HttpMiddlewareOptions = {
        host: `https://api.${REGION}.gcp.commercetools.com`,
        getAbortController: () => new AbortController(),
        timeout: 2000,
        enableRetry: true,
        retryConfig: {
          maxRetries: 2,
          retryCodes: [
            504,
            500,
            'ETIMEDOUT',
            'ECONNREFUSED',
            503,
          ],
        },
      };


      const ctpClient = new ClientBuilder()
        .withProjectKey(config.environment.COMMERCETOOLS_PROJECT_KEY)
        .withClientCredentialsFlow(authMiddlewareOptions)
        .withHttpMiddleware(httpMiddlewareOptions)
        .build();
```

ajimae:
Hi @olup,

Currently what we have with the SDK doesn't retry the request on `timeout` once a request times out we abort the request. We don't currently have a retry on aborted requests.

However, we can look into this and see if it's a feature we could implement then we will certainly implement it.

Thanks

olup:
[...]

PR #361, opened 2022-11-01
feat(http-client): add option to retry a request aborted by timeout

mh-flink:
The current behavior of the client is to retry on a timeout, but immediately abort again due to the re-use of the AbortController. This change does one adjustment and introduces a new option

By default, aborted requests will not be retried.
This is technically a behavioral change, but does not affect the outcome. The reason for this is that the logic did retry, but the http client did immediately abort.

Introduce `retryOnAbort` retry option.
The client continue retries even if it was aborted by a timeout. This is achieved by re-initializing the AbortController. The idea behind this is to allow fast retries for slow request.

Associated to Issue https://github.com/commercetools/commercetools-sdk-typescript/issues/354

### DIFF
--- a/packages/sdk-client/test/http.test/http.test.ts
+++ b/packages/sdk-client/test/http.test/http.test.ts
@@ -287,6 +287,97 @@ describe('Http', () => {
     })
   })
 
+  test('execute a get request with no retry on abort', () => {
+    expect.assertions(1)
+    return new Promise((resolve: Function, reject: Function) => {
+      const request = createTestRequest({
+        uri: '/foo/bar',
+      })
+      const response = { resolve: Function, reject: Function } as any
+
+      const next = (req: MiddlewareRequest, res: MiddlewareResponse) => {
+        expect(res).toEqual({
+          ...response,
+          error: expect.any(Error),
+          statusCode: 0,
+        })
+        resolve()
+      }
+      // Use default options
+      const httpMiddleware = createHttpMiddleware({
+        host: testHost,
+        timeout: 10, // time out after 10ms
+        fetch,
+        enableRetry: true,
+        retryConfig: {
+          retryTimes: 2,
+          retryOnAbort: false,
+        },
+        getAbortController: () => new AbortController(),
+      } as any)
+      nock(testHost)
+        .defaultReplyHeaders({
+          'Content-Type': 'application/json',
+        })
+        .get('/foo/bar')
+        .delay(100) // delay response with 100ms
+        .once()
+        .reply(200, { foo: 'bar' })
+
+      httpMiddleware(next)(request, response)
+    })
+  })
+
+  test('execute a get request with retry on abort', () => {
+    expect.assertions(1)
+    return new Promise((resolve: Function, reject: Function) => {
+      const request = createTestRequest({
+        uri: '/foo/bar',
+      })
+      const response = { resolve: Function, reject: Function } as any
+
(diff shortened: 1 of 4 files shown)


packages/sdk-client/test/http.test/http.test.ts
```diff
@@ -287,6 +287,97 @@ describe('Http', () => {
...
+        .reply(200, { foo: 'bar' })
+
+      nock(testHost)
+        .defaultReplyHeaders({
+          'Content-Type': 'application/json',
+        })
+        .get('/foo/bar')
+        .delay(10) // delay response with 100ms
+        .once()
+        .reply(200, { foo: 'bar' })
```
ajimae, line 375: What are we you trying to achieve by calling the this twice?
If it's to simulate the retry then we can collapse them into a single call by doing this:

```suggestion
      nock(testHost)
        .defaultReplyHeaders({
          'Content-Type': 'application/json',
        })
        .get('/foo/bar')
        .delay(100) // delay response with 100ms
        .times(2)
        .reply(200, { foo: 'bar' })
```

packages/sdk-client/test/http.test/http.test.ts
```diff
@@ -287,6 +287,97 @@ describe('Http', () => {
     })
   })
 
+  test('execute a get request with no retry on abort', () => {
+    expect.assertions(1)
```
ajimae, line 291:
```suggestion
    expect.assertions(2)
```

packages/sdk-client/test/http.test/http.test.ts
```diff
@@ -287,6 +287,97 @@ describe('Http', () => {
...
+        resolve()
+      }
+      // Use default options
+      const httpMiddleware = createHttpMiddleware({
+        host: testHost,
+        timeout: 10, // time out after 10ms
+        fetch,
+        enableRetry: true,
+        retryConfig: {
+          retryTimes: 2,
```
ajimae, line 313:
```suggestion
          maxRetries: 2,
```

packages/sdk-client/test/http.test/http.test.ts
```diff
@@ -287,6 +287,97 @@ describe('Http', () => {
...
+        resolve()
+      }
+      // Use default options
+      const httpMiddleware = createHttpMiddleware({
+        host: testHost,
+        timeout: 50, // time out after 10ms
+        fetch,
+        enableRetry: true,
+        retryConfig: {
+          retryTimes: 2,
```
ajimae, line 354:
```suggestion
          maxRetries: 2,
```

packages/sdk-client/test/http.test/http.test.ts
```diff
@@ -287,6 +287,97 @@ describe('Http', () => {
...
+  test('execute a get request with no retry on abort', () => {
+    expect.assertions(1)
+    return new Promise((resolve: Function, reject: Function) => {
+      const request = createTestRequest({
+        uri: '/foo/bar',
+      })
+      const response = { resolve: Function, reject: Function } as any
+
+      const next = (req: MiddlewareRequest, res: MiddlewareResponse) => {
+        expect(res).toEqual({
```
ajimae, line 299:
```suggestion
        expect(res.error.retryCount).toEqual(2)
        expect(res).toEqual({
```

ajimae (approved):
LGTM!

PR #474, opened 2023-05-10
Devx 52 integration tests examples

barbara79:
- Category
- Cart
- Cart Discount
- Channel
- Custom Object
- Customer
   - Customer => customerClone is testing mixin file -> not applicable here
- Customer Group
- Discount Code
- Extension
- GraphQL
- Inventory
- Message
- Payment
- Product
- Product Discount
- Product Projection
- Product Type
- Project
- Review
- Shipping Method
- Shopping List
- State
- Store
- Subscription
- Tax Category
- Type
- Zone

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,12 +95,12 @@ jobs:
       - run:
           name: Installing dependencies
           command: yarn install --frozen-lockfile
-      - run:
-          name: Send code coverage report to Codecov.io
-          command: yarn coverage || true
 #      - run:
-#          name: Merge coverage reports and upload to Codecov
-#          command: yarn coverage -F build_and_test,integration_test
+#          name: Send code coverage report to Codecov.io
+#          command: yarn coverage || true
+      - run:
+          name: Merge coverage reports and upload to Codecov
+          command: yarn coverage -F build_and_test,integration_test
 
 #  update_ml_sdk_from_spec:
 #    executor: << parameters.os >>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -95,9 +95,9 @@ jobs:
       - run:
           name: Installing dependencies
           command: yarn install --frozen-lockfile
-      - run:
-          name: Send code coverage report to Codecov.io
-          command: yarn coverage || true
+#      - run:
+#          name: Send code coverage report to Codecov.io
+#          command: yarn coverage || true
       - run:
           name: Merge coverage reports and upload to Codecov
           command: yarn coverage -F build_and_test,integration_test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,6 +23,8 @@ executors:
 version: 2.1
 orbs:
   node: circleci/node@5.1.0
+  codecov: codecov/codecov@3.8.3
+
 jobs:
   build_and_test:
     parameters:
@@ -89,6 +91,7 @@ jobs:
     working_directory: *working_directory
     steps:
       - checkout
+      - setup_remote_docker
       - restore_cache: *restore_yarn_cache
       - run:
           name: Merge coverage reports
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,10 @@ jobs:
         type: string
     executor: << parameters.os >>
     working_directory: *working_directory
+    # environment:
+      # CTP_PROJECT_KEY: $CTP_PROJECT_KEY
+      # CTP_CLIENT_ID: $CTP_CLIENT_ID
+      # CTP_CLIENT_SECRET: $CTP_CLIENT_SECRET
     steps:
       - checkout
       - restore_cache: *restore_yarn_cache
@@ -41,6 +45,9 @@ jobs:
           name: Installing dependencies
           command: yarn install --frozen-lockfile
       - save_cache: *save_yarn_cache
+      - run: 
+          name: Echo env
+          command: sh -c env
       - run:
           name: Install example dependencies
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,7 +97,7 @@ jobs:
           command: yarn install --frozen-lockfile
       - run:
           name: Upload coverage report
-          command: codecov
+          command: yarn coverage
 
 #  update_ml_sdk_from_spec:
 #    executor: << parameters.os >>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -99,11 +99,11 @@ jobs:
           name: Create coverage directory
           command: mkdir coverage
       - run:
-          name: Install Codecov Bash uploader
-          command:  codecov
+          name: Send code coverage report to Codecov.io
+          command: yarn coverage || true
       - run:
           name: Merge coverage reports and upload to Codecov
-          command: codecov -Z -F build_and_test -F integration_test
+          command: yarn codecov -Z -F build_and_test -F integration_test
 
 #  update_ml_sdk_from_spec:
 #    executor: << parameters.os >>
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,7 +60,7 @@ jobs:
 #      - store_artifacts:
 #          path: coverage
       - codecov/upload:
-          file: {{ coverage_report_filepath }}
+          file: coverage/clover.xml
 #  integration_test:
 #    parameters:
 #      os:
@@ -137,14 +137,14 @@ workflows:
               os: [docker]
 #              node-version: ["^14.20.0", "^16.17.0", "^18.9.0"]
               node-version: ["^18.9.0"]
-      - integration_test:
-          requires:
-            - build_and_test
-          matrix:
-            parameters:
-              os: [ docker ]
-              node-version: ["^18.9.0"]
-#      - coverage_report:
+#      - integration_test:
+#          requires:
+#            - build_and_test
+#          matrix:
+#            parameters:
+#              os: [ docker ]
+#              node-version: ["^18.9.0"]
+##      - coverage_report:
 #          requires:
 #            - build_and_test
 #            - integration_test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -97,7 +97,7 @@ jobs:
           command: yarn install --frozen-lockfile
       - run:
           name: Install lcov-result-merger
-          command: yarn add --dev lcov-result-merger
+          command: yarn add --dev -W lcov-result-merger
       - run:
           name: Create coverage directory
           command: mkdir coverage
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -61,6 +61,7 @@ jobs:
 #          path: coverage
       - codecov/upload:
           file: coverage/clover.xml
+          flags: unittests
   integration_test:
     parameters:
       os:
@@ -83,6 +84,7 @@ jobs:
 #          path: coverage
       - codecov/upload:
           file: coverage/clover.xml
+          flags: integrationtests
 #  coverage_report:
 #    parameters:
 #      os:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,7 +133,8 @@ workflows:
           matrix:
             parameters:
               os: [docker]
-              node-version: ["^14.20.0", "^16.17.0", "^18.9.0"]
+#              node-version: ["^14.20.0", "^16.17.0", "^18.9.0"]
+              node-version: ["^18.9.0"]
       - integration_test:
           requires:
             - build_and_test
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,17 +55,38 @@ jobs:
       - run:
           name: Running tests
           command: yarn test --testPathPattern='packages/platform-sdk/test/(?!integration-tests).*' --coverage
+      - store_artifacts:
+          path: coverage
+
+  integration_test:
+    parameters:
+      os:
+        type: executor
+      node-version:
+        type: string
+    executor: << parameters.os >>
+    working_directory: *working_directory
+    steps:
+      - checkout
+      - restore_cache: *restore_yarn_cache
       - run:
-          name: Send coverage report
+          name: Installing dependencies
+          command: yarn install --frozen-lockfile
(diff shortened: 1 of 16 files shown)


lojzatran (commented):
IMO this PR is a good start 👍  but I do not see in the JIRA task nor here what is the goal of this task. How many tests do you want to add and based on what you will consider which tests would be added?

packages/platform-sdk/test/integrationTests/api-client/api-client-fixture.ts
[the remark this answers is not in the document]
barbara79: I will remove it for now but I will add it again once I will write the tests for the ApiClient

packages/platform-sdk/test/integrationTests/customer/customer.test.ts
```diff
@@ -0,0 +1,210 @@
...
+      })
+      .execute()
+
+    expect(responseCustomerDeleted.statusCode).toEqual(200)
+
+    await deleteCustomerGroup(customerGroup)
+  })
+
+  it('should update a customer by Id', async () => {
+    const customerGroup = await createCustomerGroup()
```
barbara79, line 66: Leaving these methods here to show the customer step-by-step how to create/delete the final object

packages/platform-sdk/test/integrationTests/discount-code/discount-code.test.ts
```diff
@@ -0,0 +1,120 @@
...
+      .execute()
+    expect(getDiscountCode).not.toBe(null)
+    expect(getDiscountCode.body.id).toEqual(discountCode.body.id)
+
+    await deleteDiscountCode(discountCode)
+    await deleteCartDiscount(cartDiscount)
+  })
+
+  it('should query a discount code', async () => {
+    const cartDiscount = await createCartDiscount()
```
barbara79, line 73: here is the same like previous comment: "Leaving these methods here to show the customer step-by-step how to create/delete the final object"

packages/platform-sdk/test/integrationTests/inventory/inventory.test.ts
```diff
@@ -0,0 +1,108 @@
...
+      })
+      .execute()
+
+    expect(responseInventoryDeleted.statusCode).toEqual(200)
+
+    await deleteChannel(channel)
+  })
+
+  it('should get an inventory by ID', async () => {
+    const channel = await createChannel()
```
barbara79, line 47: here is the same like previous comment: "Leaving these methods here to show the customer step-by-step how to create/delete the final object"

packages/platform-sdk/test/integrationTests/shopping-list/shopping-list.test.ts
```diff
@@ -0,0 +1,223 @@
...
+} from '../customer-group/customer-group-fixture'
+import {
+  createShoppingList,
+  createShoppingListDraft,
+  deleteShoppingList,
+} from './shopping-list-fixture'
+
+describe('testing shopping list API calls', () => {
+  it('should create and delete a shopping list by ID', async () => {
+    const customerGroup = await createCustomerGroup()
```
barbara79, line 26: here is the same like previous comment: "Leaving these methods here to show the customer step-by-step how to create/delete the final object"

packages/platform-sdk/test/integrationTests/product-projection/product-projection.test.ts
```diff
@@ -0,0 +1,315 @@
...
+    expect(productProjectionSearchResponse.body.facets).not.toBe(null)
+
+    await deleteProduct(product)
+    await deleteProductType(productType)
+    await deleteTaxCategory(taxCategory)
+    await deleteCategory(category)
+  })
+
+  // it('should get a product by SKU using query predicates', async () => {
+  //   const category = await createCategory()
```
evansinho, line 176: Why do we have this commented code?

evansinho (commented):
Apart from the commented code, it looks fine

packages/platform-sdk/test/integrationTests/product-projection/product-projection.test.ts
barbara79: you are right :D, I already removed them in my locale I will clean them up in the next push ;)

packages/platform-sdk/test/integrationTests/test-utils.ts
```diff
@@ -0,0 +1,30 @@
+import { ClientBuilder } from '@commercetools/sdk-client-v2/src'
```
ajimae, line 1: This should fix the `invalid token error`

```suggestion
import { ClientBuilder } from '@commercetools/sdk-client-v2'
```

barbara79:
> Also, add a before script to clear all data before running these tests.

where should I put it? every tests calls common endpoints

barbara79:
> we should have a `before()` or `afterAll()` to remove all the data produced by the test.

I think I have an idea about how to do it:
I'll create a file for the cancelation of the data and then create methods for each endpoint and call them in a `beforeAll` in each test file

.circleci/config.yml
```diff
@@ -95,6 +117,11 @@ workflows:
   validate_spec:
     jobs:
       - build_and_test:
+          matrix:
+            parameters:
+              os: [docker]
+              node-version: ["^14.20.0", "^16.17.0", "^18.9.0"]
```
ajimae, line 123: You don't need to create all these matrices for the integration test, only one node_version is sufficient.

ajimae (commented):
I think we should use just one node_version for the integration test.

.circleci/config.yml
barbara79: I updated see line 128.

barbara79:
> > we should have a `before()` or `afterAll()` to remove all the data produced by the test.
> 
> I think I have an idea about how to do it: I'll create a file for the cancelation of the data and then create methods for each endpoint and call them in a `beforeAll` in each test file

we decided in the Weekly that the delete methods that are already for each test are enough, so no need to add before or after methods

.circleci/config.yml
```diff
@@ -66,6 +66,37 @@ jobs:
...
+      - run:
+          name: Installing dependencies
+          command: yarn install --frozen-lockfile
+      - save_cache: *save_yarn_cache
+      - run:
+          name: Use environment variables
+          command: |
+            echo $IT_CTP_PROJECT_KEY
+            echo $IT_CTP_CLIENT_SECRET
+            echo $IT_CTP_CLIENT_ID
```
ajimae, line 89: I don't see the need for this step, I believe we can remove it now.

.circleci/config.yml
```diff
@@ -54,7 +54,7 @@ jobs:
             yarn test
       - run:
           name: Running tests
-          command: yarn test --coverage
+          command: yarn test --coverage --testPathPattern='packages/platform-sdk/test/(?!integration-tests).*'
```
ajimae, line 57: This change should be reverted to what it was.

```suggestion
          command: yarn test --coverage
```

.circleci/config.yml
```diff
@@ -66,6 +66,37 @@ jobs:
...
+      - save_cache: *save_yarn_cache
+      - run:
+          name: Use environment variables
+          command: |
+            echo $IT_CTP_PROJECT_KEY
+            echo $IT_CTP_CLIENT_SECRET
+            echo $IT_CTP_CLIENT_ID
+      - run:
+          name: Running integration tests
+          command: yarn test --coverage --testPathPattern='packages/platform-sdk/test/integration-tests'
```
ajimae, line 92: Test coverages shouldn't be test specific but as a collection of the whole tests suites, even though the tests runs one at a time, the coverage should be an aggregate of all the tests.

ajimae (commented):
Nice work 🎉

However, I believe the test coverage can still be improved as it stands now, the coverage will be degraded to 7% from the original value of 98%

Thanks

barbara79:
> Nice work 🎉
> 
> However, I believe the test coverage can still be improved as it stands now, the coverage will be degraded to 7% from the original value of 98%
> 
> Thanks

Thanks, that's because I'm trying to combine the tests.
Before it was checking only the unit tests and now I'm adding the integration tests, but since there are 2 jobs separated it doesn't count them very well.
I'm on it anyway

.circleci/config.yml
barbara79: it's to pass the variable for the integration tests otherwise it doesn't catch them

.circleci/config.yml
barbara79: it's still WIP
I'm trying to find a way to combine the coverage for both kind of tests

barbara79: I'll let you know when you can review again the PR

.circleci/config.yml
barbara79: and it runs everything in a specific environment which is what we wanted to do

barbara79:
For Codecov created another task to check the reason why it decreased (see the Epic)
https://app.codecov.io/gh/commercetools/commercetools-sdk-typescript/tree/DEVX-52-integration-tests-examples

packages/platform-sdk/test/integration-tests/order/order.test.ts
```diff
@@ -0,0 +1,227 @@
...
+    await deleteProduct(product)
+    await deleteProductType(productType)
+    await deleteTaxCategory(taxCategory)
+    await deleteCategory(category)
+  })
+
+  it('should search a order', async () => {
+    let project = await ctpApiBuilder.get().execute()
+
+    if (project.body.searchIndexing.orders.status === 'Activated') {
```
lojzatran, line 150: It's better to ensure that the project has order search activated than reactivate it every time.

So if the status is `Activated`, you can skip `changeOrderSearchStatus` action.

packages/platform-sdk/test/integration-tests/order/order.test.ts
```diff
@@ -0,0 +1,227 @@
...
+      .get()
+      .execute()
+    await deleteCart(getCart)
+    await deleteProduct(product)
+    await deleteProductType(productType)
+    await deleteTaxCategory(taxCategory)
+    await deleteCategory(category)
+  })
+
+  it('should search a order', async () => {
```
lojzatran, line 147: I think it's better to disable this test with a comment why. Right now this test is a bit confusing, because it does not do the search, even though the name says so.

packages/platform-sdk/test/integration-tests/product-projection/product-projection.test.ts
```diff
@@ -0,0 +1,177 @@
...
+      .search()
+      .get({
+        queryArgs: {
+          facet: 'categories.id',
+          staged: true,
+        },
+      })
+      .execute()
+    expect(productProjectionSearchResponse).not.toBe(null)
+    expect(productProjectionSearchResponse.body.facets).not.toBe(null)
```
lojzatran, line 170: This test might be flaky since search indexing is required and it takes a while. Also it does not ensure that the search indexing is activated for the project.

lojzatran (commented):
Good job 👍

packages/platform-sdk/test/integration-tests/order/order.test.ts
barbara79: you are right! I don't know why I did this XD

packages/platform-sdk/test/integration-tests/product-projection/product-projection.test.ts
barbara79: I would also put this test in the task that I opened since I don't have a lot of time to fix everything 
I'm sorry :(

barbara79: this is the task opened https://commercetools.atlassian.net/browse/DEVX-173

packages/platform-sdk/test/integration-tests/custom-object/custom-object.test.ts
```diff
@@ -0,0 +1,143 @@
...
+    const queryCustomObject = await apiRoot
+      .customObjects()
+      .withContainerAndKey({
+        container: customObject.body.container,
+        key: customObject.body.key,
+      })
+      .get({
+        queryArgs: {
+          where: 'id=' + '"' + customObject.body.id + '"',
+        },
```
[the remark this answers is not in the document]
lojzatran: I think you should change this to another example, because it looks like query by ID like this is OK too and actually it is not. For custom object you could query by `createdAt`.

packages/platform-sdk/test/integration-tests/cart-discount/cart-discount.test.ts
```diff
@@ -0,0 +1,187 @@
...
+  })
+
+  it('should query a cart discount', async () => {
+    const cartDiscount = await createCartDiscount()
+
+    const queryCartDiscount = await apiRoot
+      .cartDiscounts()
+      .get({
+        queryArgs: {
+          where: 'id=' + '"' + cartDiscount.body.id + '"',
```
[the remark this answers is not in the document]
lojzatran: https://github.com/commercetools/commercetools-sdk-typescript/pull/474/commits/9d93a234b5478a9071c6de0666170b1c0e32c8b0

packages/platform-sdk/test/integration-tests/graphql/graphql.test.ts
```diff
@@ -0,0 +1,58 @@
...
+      query,
+    }
+    const graphQLResponse = await apiRoot
+      .graphql()
+      .post({
+        body: graphQLRequest,
+      })
+      .execute()
+
+    expect(graphQLResponse).not.toBe(null)
```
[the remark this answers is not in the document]
barbara79: I'm gonna add it this task https://commercetools.atlassian.net/jira/software/c/projects/DEVX/boards/386?modal=detail&selectedIssue=DEVX-173

packages/platform-sdk/test/integration-tests/custom-object/custom-object.test.ts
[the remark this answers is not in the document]
barbara79: https://github.com/commercetools/commercetools-sdk-typescript/pull/474/commits/30df75a296b20fa7b74e44263eed3e70bd141d0c

evansinho (approved):

PR #493, opened 2023-06-20
Update SUNRISE data link

industrian:
We will be removing the SUNRISE data page from the docs soon, so I'm changing the link to point to the GitHub repo.


ajimae (approved):
LGTM!

PR #531, opened 2023-09-28
Refactor to not use long-lived PAT but short-lived PAT issued by GitHub App

tdeekens:
#### Summary

Replaces the usage of a static long-lived PAT with a short-lived (auto revoked) PAT through a GitHub Application called CT Changesets.

This is being rolled out across about 10 repositories. More detail and background on the changes can be found on internal documentation.

### DIFF
--- a/.github/workflows/create-pr.yml
+++ b/.github/workflows/create-pr.yml
@@ -11,11 +11,19 @@ jobs:
 
     runs-on: ubuntu-latest
     steps:
+      # Get GitHub token via the CT Changesets App
+      - name: Generate GitHub token (via CT Changesets App)
+        id: generate_github_token
+        uses: tibdex/github-app-token@v2.1.0
+        with:
+          app_id: ${{ secrets.CT_CHANGESETS_APP_ID }}
+          private_key: ${{ secrets.CT_CHANGESETS_APP_PEM }}
+          
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


.github/workflows/create-pr.yml
```diff
@@ -11,11 +11,19 @@ jobs:
...
+          app_id: ${{ secrets.CT_CHANGESETS_APP_ID }}
+          private_key: ${{ secrets.CT_CHANGESETS_APP_PEM }}
+          
       - name: Checkout
         uses: actions/checkout@v3
       - uses: jenschude/auto-create-pr-action@v0.3.1
         with:
           request_title: "Update generated SDKs"
-          github_token: ${{ secrets.GITHUB_TOKEN }}
+          github_token: ${{ steps.generate_github_token.outputs.token }}
```
emmenko, line 27: Do you know what token is this? It's called `GITHUB_TOKEN` and not `RELEASE_GITHUB_TOKEN` 🤔

tdeekens: It doesn't exist on the repo as a secret. It's the "default" token. Maybe we don't have to remove it then 🤔 

https://docs.github.com/en/actions/security-guides/automatic-token-authentication

WDYT?

emmenko: Ah ok, then I would leave it as-is

tdeekens: Done

emmenko (approved):

PR #545, opened 2023-10-16
Fix User Agent Header 

ajimae:
### Summary
Append user agent from header if exists

### Tasks completed
- [x] Check if `User-Agent` header already exists, then append it.

### DIFF
--- a/packages/sdk-client/src/client-builder/ClientBuilder.ts
+++ b/packages/sdk-client/src/client-builder/ClientBuilder.ts
@@ -198,7 +198,12 @@ export default class ClientBuilder {
   withTelemetryMiddleware<T extends TelemetryOptions<T>>(
     options: T
   ): ClientBuilder {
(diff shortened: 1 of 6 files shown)


packages/sdk-client/src/client-builder/ClientBuilder.ts
```diff
@@ -198,7 +198,12 @@ export default class ClientBuilder {
   withTelemetryMiddleware<T extends TelemetryOptions<T>>(
     options: T
   ): ClientBuilder {
+    const { version } = require('@commercetools/ts-sdk-apm/package.json')
     const { createTelemetryMiddleware, ...rest } = options
+
+    this.withUserAgentMiddleware({
+      customAgent: `typescript-sdk-newrelic-middleware/${version}`,
```
jenschude, line 205: Nice! Now it becomes a static string and also uses the composition of the the different agent parts.

Also this may override custom agent name from a user we should go with this for now

jenschude (approved):

PR #617, opened 2024-02-23
Fix concurrent modification middleware

lojzatran:
The issue was in `client.ts`. It was both resolving/rejecting the promise + returning the response. Removing resolve/reject the promise and instead throwing the error or returning the response helped.
My explanation to this problem is that rejecting the promise fulfils the promise. However, this promise is resolved in the `execute` method, therefore it skips all the middlewares in between.


packages/sdk-client-v3/tests/client.test/client.test.ts
```diff
@@ -212,69 +212,6 @@ describe('execute function', () => {
         return Promise.resolve()
       })
   })
-
```
ajimae, line 215: Why are we removing this tests? 
I think we should have rewritten this test to match the new client implementation.

packages/sdk-client-v3/tests/concurrent-modification.test/concurrent-modification-middleware.test.ts
```diff
@@ -39,7 +39,7 @@ describe('Concurrent Modification Middleware.', () => {
     const request = createTestRequest({ body: { version: 4 } })
     const response = createTestResponse({
       statusCode: 409,
-      error: { body: { errors: [{ currentVersion: 5 }] } },
```
ajimae, line 42: Shouldn't this be an error response?

packages/sdk-client-v3/src/client/client.ts
```diff
@@ -162,11 +162,9 @@ export default function createClient(middlewares: ClientOptions): Client {
       } as MiddlewareResponse
 
       if (res.error) {
-        res.reject(res.error)
```
ajimae, line 165: Don't you think it will be better to return the error here instead of throwing it, so that the ErrorMiddleware can properly construct the error based on this returned value?

lojzatran: https://github.com/commercetools/commercetools-sdk-typescript/pull/617/commits/82ef8ac299dea3ed4e3f7c7e5184e34213524472

packages/sdk-client-v3/tests/concurrent-modification.test/concurrent-modification-middleware.test.ts
lojzatran: https://github.com/commercetools/commercetools-sdk-typescript/pull/617/commits/82ef8ac299dea3ed4e3f7c7e5184e34213524472#diff-51569edfbd3a5b4db84c8b79a0bd1cb94ebc092a0ebdd27387d15b683f310215R40

ajimae (approved):

PR #621, opened 2024-02-27
Fixing Wrong Type Information in sdk-client-v3

Engerim:
- the type MiddlewareResponse states that the field is named request, but it is originalRequest
- adjust the logger middleware to use original request field


ajimae (approved):

Engerim:
any update on that as the wrong type information still consists and so the logger is not behaving correctly

PR #637, opened 2024-03-13
Fix Client Process Function

ajimae:
### Summary
Fix Client Process Function

### Tasks Completed
- [x] fix static limit for the process fn options
- [x] remove old implementations to use limit from options


packages/sdk-client/src/sdk-client/client.ts
```diff
@@ -58,13 +60,13 @@ export function process(
...
       // defaults
-      limit: 20,
+      limit: opt.limit,
       // merge given query params
       ...requestQuery,
     }
 
-    let hasFirstPageBeenProcessed = false
     let itemsToGet = opt.total
+    let hasFirstPageBeenProcessed = false
```
lojzatran, line 69: Why this change?

packages/sdk-client/src/utils/index.ts
```diff
@@ -1 +1,2 @@
 export { parseURLString, stringifyURLString } from './url';
+export const PAGE_LIMIT = 20;
```
lojzatran, line 2: It‘s better to keep this in `client.ts` file since it‘s used only there.

packages/sdk-client/test/client.test/sdk-client.test.ts
```diff
@@ -8,7 +8,7 @@ import {
 import fetch from 'node-fetch'
```
lojzatran, line 8: Please add a test that proves your fix is actually working - a test that fails on master branch but passes on your branch.

lojzatran (changes requested):
In the JIRA task you wrote that `sort` has hardcoded value, but here sort is not changed anywhere. And in the github issue the cusomer was reporting an error related to expansion and this is also not addressed here. Can you explain why?

ajimae:
> In the JIRA task you wrote that `sort` has hardcoded value, but here sort is not changed anywhere. And in the github issue the cusomer was reporting an error related to expansion and this is also not addressed here. Can you explain why?

This is obviously a mistake. I have corrected the JIRA task.

packages/platform-sdk/test/integration-tests/process/process.test.ts
```diff
@@ -0,0 +1,153 @@
...
+
+    const limit = 30
+    const fn = (data: any) => data
+
+    client
+      .process(request, fn, { limit })
+      /**
+       * response is an array of processed results
+       */
+      .then((response) => {
```
lojzatran, line 47: It‘s better to use await here since you defined already the async function.

lojzatran (approved):
Good job! 👍

PR #453, opened 2023-03-26
Remove all `querystring` dependency

ajimae:
### Summary

- [x] replace all occurrence of deprecated `querystring` dependency with `qs`
- [x] add `qs` dependency to required packages
- [x] refactor code to accommodate new implementation

### Devs Note:
This [PR](https://github.com/commercetools/rmf-codegen/pull/301) need to be merged before this. This is to enable failing tests to be fix in this PR.

### Related Issue
#452

PR #671, opened 2024-04-08
Fix malformed uri parameter error

### LINKED ISSUE #513: client.process fails with "expand: Invalid path selector"
Request:
`const request = api
  .categories()
  .get({
    queryArgs: {
      limit: 500,
      withTotal: false,
      expand: ['parent', 'ancestors[*]'],
    },
  })
  .clientRequest();`
  
  `client.process(request, ....)` fails with message "Malformed parameter: expand: Invalid path selector: parent,ancestors[*]"}]},"name":"BadRequest"},"message":"Malformed parameter: expand: Invalid path selector: parent,ancestors[*]"
  
  i think this bug was introduced in #453
  v2.1.5 is working fine.

ajimae:
### Summary
Fixes a bug that prevents query args from being properly formatted.

### Completed Tasks
- [x] fix the bug introduced by #453 PR by using the proper stringifer
- [x] use the proper javascript serialization function
- [x] add tests to validate functionality and the fix
- [x] add release changesets

### Fixes
#513

### DIFF
--- a/packages/platform-sdk/test/integration-tests/process/process.test.ts
+++ b/packages/platform-sdk/test/integration-tests/process/process.test.ts
@@ -123,7 +123,7 @@ describe('integration test for process function', () => {
       .catch(fn)
   })
 
-  test('should properly the `expan` args as a uri parameter', () => {
+  test('should properly the `expand` args as a uri parameter', () => {
     const client = new ClientBuilder()
       .withProjectKey(projectKey)
       .withClientCredentialsFlow(getAuthOptions())


packages/platform-sdk/test/integration-tests/process/process.test.ts
```diff
@@ -150,4 +122,34 @@ describe('integration test for process function', () => {
       })
       .catch(fn)
   })
+
+  test('should properly the `expan` args as a uri parameter', () => {
```
lojzatran, line 126:
```suggestion
  test('should properly the `expand` args as a uri parameter', () => {
```

packages/platform-sdk/test/integration-tests/process/process.test.ts
```diff
@@ -150,4 +122,34 @@ describe('integration test for process function', () => {
...
+          expand: ['parent', 'ancestors[*]'],
+        },
+      })
+      .clientRequest()
+
+    const fn = (data: any) => data
+    return client
+      .process(request, fn, {})
+      .then((response) => {
+        expect(response[0].statusCode).toEqual(200)
```
lojzatran, line 151: I tried to run this test also in master and it is passing as well, so it does not show that the error is fixed or not. Please fix this IT so that it fails on master branch and not in your branch to show that the error is fixed.

lojzatran (changes requested):

packages/platform-sdk/test/integration-tests/process/process.test.ts
ajimae: I don't know what test you ran but the issue is on master, there is nothing wrong with the test that needs fixing.

Here is the same issue in the master branch.


ezgif-7-ea06c3391e

ajimae: And this is a commit before the bug was introduced `d06e0c5`. 

Same code snippet ran with the commit can be seen in the attached clip and it doesn't have the said bug.

I suggest you properly test this from your end.

If you need an help with this let me know.

Thank you.

https://github.com/commercetools/commercetools-sdk-typescript/assets/32770340/cbd6d58d-22ca-43c3-95ba-8f171c899166

lojzatran: Thanks for the snippets. I tried again and the test is still passing. My procedure is following:
1. I copied only the test `should properly the `expan` args as a uri parameter` to master branch.
2. I ran the command `npm test -- process.test.ts` in terminal.
3. The tests are passing and I would expect one test to fail.

I see in the snippet you don‘t do the same thing. Can you try what I did?

lojzatran (approved):
Good job! 🚀

Issue #647, opened 2024-03-23
@commercetools/ts-client 1.2.0 not throwing errors anymore


mvantellingen:
It seems that the `@commercetools/ts-client@1.2.0` is not throwing errors anymore when the server returns an error. Might be related to #620

PR #678, opened 2024-04-13
Fix error not thrown

lojzatran:
Fixes: https://github.com/commercetools/commercetools-sdk-typescript/issues/647

@ajimae : I tried to fail the test in case the error is not thrown, but `fail()` is not supported by jest: https://github.com/jestjs/jest/issues/11698. Do you know how to do this?


ajimae:
> Fixes: #647
> 
> @ajimae : I tried to fail the test in case the error is not thrown, but `fail()` is not supported by jest: [jestjs/jest#11698](https://github.com/jestjs/jest/issues/11698). Do you know how to do this?

So I tried it locally, the error is not being throw, it is being returned as a response as well, instead of being rejected and passed to the the `.catch()` block.

lojzatran:
> > Fixes: #647
> > @ajimae : I tried to fail the test in case the error is not thrown, but `fail()` is not supported by jest: [jestjs/jest#11698](https://github.com/jestjs/jest/issues/11698). Do you know how to do this?
> 
> So I tried it locally, the error is not being throw, it is being returned as a response as well, instead of being rejected and passed to the the `.catch()` block.

Can you provide more details which error is that so I can write another test for it? I tested for 404 HTTP error and it is now throwing.

packages/platform-sdk/test/integration-tests/sdk-v3/error-cases.test.ts
```diff
@@ -0,0 +1,17 @@
...
+describe('testing error cases', () => {
+  it('should throw error when a product type is not found', async () => {
+    try {
+      await apiRootV3
+        .productTypes()
+        .withId({ ID: 'non-existing-id' })
+        .get()
+        .execute()
+
+      throw new Error('Should have thrown an error')
```
ajimae, line 12: Indeed it does work for me now, however I think this line shouldn't be here, the sdk call chain should be the one throwing the error when the `.execute()` is called, then we can process the error and make our checks in the `.catch()` block.

So aparently, try-catch will not work here because of the async nature of the builder chain, you might want to remove the try-catch and handle the error in the `.cache()`.


```ts
  it('should throw error when a product type is not found', async () => {
    await apiRootV3
      .productTypes()
      .withId({ ID: 'non-existing-id' })
      .get()
      .execute()
      .catch(e => {
        console.log(e, '---------------->')
        expect(e.statusCode).toEqual(404)
      })
  })
```

I tried the above and it worked just fine. So we can restructure the test a bit to include the catch block instead of throwing a generic error in the `try-catch` setup.

ajimae (commented):
@lojzatran this is my finding. What do you think?

packages/platform-sdk/test/integration-tests/sdk-v3/error-cases.test.ts
lojzatran: Thanks for the suggestion. I checked it and there is one problem in your test: if the error is not thrown (the code before my changes), then the test would also pass. This means that it does not test my fix.

Here I explained it a bit better:

```js
it('should throw error when a product type is not found', async () => {
    await apiRootV3
      .productTypes()
      .withId({ ID: 'non-existing-id' })
      .get()
      .execute()
      .then(() => {
          console.log('it should throw an error when ID is non-existing and not reach this console.log')
          fail()  // <--- a method to fail the test
      })
      .catch(e => {
        expect(e.statusCode).toEqual(404)
      })
  })
```

There is usually `fail()` method to fail the test if it reaches a place it shouldn't reach, but it seems to be missing in jest. If you didn‘t encounter it before, it‘s ok and I will leave the test as it is.

ajimae: Since this is an integrated test, I want to believe that the test must fail, since the ID: 'non-existing-id' is actually a non existent ID.

ajimae (approved):

ajimae:
> ### ⚠️ No Changeset found
> Latest commit: [a90c14b](https://github.com/commercetools/commercetools-sdk-typescript/commit/a90c14b70f016b3bb12b6647373f2f06191519b0)
> 
> Merging this PR will not cause a version bump for any packages. If these changes should not result in a new version, you're good to go. **If these changes should result in a version bump, you need to add a changeset.**
> 
> This PR includes no changesets
> [Click here to learn what changesets are, and how to add one](https://github.com/changesets/changesets/blob/main/docs/adding-a-changeset.md).
> 
> [Click here if you're a maintainer who wants to add a changeset to this PR](https://github.com/commercetools/commercetools-sdk-typescript/new/647_ts-client_not_throwing_error?filename=.changeset/rude-hairs-own.md&value=---%0A%22%40commercetools%2Fplatform-sdk%22%3A%20patch%0A%22%40commercetools%2Fts-client%22%3A%20patch%0A%22%40commercetools%2Fsdk-client-v2%22%3A%20patch%0A---%0A%0AFix%20error%20not%20thrown%0A)

We didn't add a changeset to release this fix. I will go ahead and do that now.

PR #705, opened 2024-05-13
[Chore] Fix Security Vulnerability

ajimae:
### Summary
Fix remaining security vulnerabilities

### Completed Tasks
- [x] add updated packages to resolution block

### DIFF
--- a/examples/react-webpack/package.json
+++ b/examples/react-webpack/package.json
@@ -39,9 +39,9 @@
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "resolutions": {
-    "semver": "7.6.0"
+    "semver": "7.5.2"
   },
   "overrides": {
-    "semver": "7.6.0"
+    "semver": "7.5.2"
   }
 }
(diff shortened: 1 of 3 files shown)


examples/react-webpack/package.json
```diff
@@ -39,9 +39,9 @@
     "test": "echo \"Error: no test specified\" && exit 1"
   },
   "resolutions": {
-    "semver": "7.6.0"
+    "semver": "7.5.2"
```
lojzatran, line 42: Why did you lower the version?

lojzatran (approved):

examples/react-webpack/package.json
jenschude, line 42: Is this a typo? Latest version is 7.6.2

ajimae: For some reason it was still complaining on version 7.6.2 [here](https://github.com/commercetools/commercetools-sdk-typescript/security/dependabot/141)

jenschude: Looking at the report seems that we have multiple resolutions for semver in the lock file

https://github.com/commercetools/commercetools-sdk-typescript/blob/ae315373964a2daba2186a38b795f708b4dc4058/examples/react-webpack/yarn.lock#L4344-L4359

ajimae: I have updated the sermver version to the latest, let see if it fixes it.

PR #714, opened 2024-05-23
feat(abortController): add retry on abort

lojzatran:
I added back this feature as it is already on v2. Additionally I added some tests as well.

And I also updated the docs: https://github.com/commercetools/commercetools-docs/pull/4986


packages/sdk-client-v3/src/utils/validate.ts
```diff
@@ -19,11 +19,6 @@ export function validateHttpOptions(options: HttpMiddlewareOptions) {
     throw new Error(
       'An `httpClient` is not available, please pass in a `fetch` or `axios` instance as an option or have them globally available.'
     )
-
-  if (options.timeout && !options.getAbortController)
```
lojzatran, line 23: See: https://github.com/commercetools/commercetools-sdk-typescript/blob/68ad0e322b355cdee28e3114c95e60d0b9191df0/packages/sdk-client-v3/src/middleware/create-http-middleware.ts#L167

ajimae:
The tests are failing.

lojzatran:
> @lojzatran the tests are failing, please make them pass so I can start the review process.

It‘s fixed now.

ajimae:
Nice work, I noticed that there is no integration tests to show (using **actual test data**) that this change actually works as intended?

packages/platform-sdk/test/integration-tests/sdk-v3/middlewares.test.ts
```diff
@@ -268,3 +268,31 @@ describe('Telemetry Middleware', () => {
     expect(telemetryMiddlewareCalled).toHaveBeenCalled()
   })
 })
+
+describe('Custom middleware', () => {
```
ajimae, line 272: Why do we have this test here, is it part of the work abort controller feature?

packages/sdk-client-v3/src/types/types.d.ts
```diff
@@ -222,7 +222,7 @@ export type HttpMiddlewareOptions = {
   retryConfig?: RetryOptions
   httpClient: Function
   getAbortController?: () => AbortController
-  httpClientOptions?: object
+  httpClientOptions?: object // will be passed as a second argument to your httpClient function for configuration
```
ajimae, line 225: I noticed that from the docs link provided [here](https://github.com/commercetools/commercetools-docs/pull/4986) there is not mention of the `httpClientOptions` as an addition to the `httpMiddlewareOption` just as we have one for `httpClient`. Please add this property to the docs.

packages/sdk-client-v3/src/types/types.d.ts
```diff
@@ -242,6 +242,7 @@ export type RetryMiddlewareOptions = {
   maxRetries?: number
   retryDelay?: number
   maxDelay?: typeof Infinity
+  retryOnAbort?: boolean
```
ajimae, line 245: Please check the docs, I could see this was removed in the v3 docs [here](https://github.com/commercetools/commercetools-docs/pull/4986/files#diff-4c9187ddfa33195642947e7508f80e1fced9d8e01127e80d1c4b2d9544f46524L16)

packages/sdk-client-v3/src/utils/executor.ts
```diff
@@ -33,13 +33,15 @@ export default async function executor(request: HttpClientConfig) {
...
-      const { enableRetry, retryConfig } = rest
+      const { enableRetry, retryConfig, abortController } = rest
       const {
         retryCodes = [],
         maxDelay = Infinity,
         maxRetries = 3,
         backoff = true,
         retryDelay = 200,
+        // If set to true reinitialize the abort controller when the timeout is reached and apply the retry config
+        retryOnAbort = false,
```
ajimae, line 44: Was there any reason why this was made `false` by default?

packages/sdk-client-v3/src/utils/executor.ts
```diff
@@ -68,17 +70,31 @@ export default async function executor(request: HttpClientConfig) {
...
+          return _response
+        }
 
+        // first attempt
+        let _response = await executeWithTryCatch(retryCodes)
         // retry attempts
-        while (enableRetry && retryCount < maxRetries) {
+        while (
+          enableRetry &&
+          (retryOnAbort || !abortController || !abortController.signal) &&
```
ajimae, line 93: What happens when `retryOnAbort` is truthy and `aborController` is falsy and vice versa?

packages/sdk-client-v3/src/utils/executor.ts
```diff
@@ -68,17 +70,31 @@ export default async function executor(request: HttpClientConfig) {
...
+        const executeWithTryCatch = async (retryCodes) => {
+          let _response = {} as any
+          try {
+            _response = await execute()
+            if (predicate(retryCodes, _response)) return _response
+          } catch (e) {
+            if (e.name.includes('AbortError')) {
+              _response = e
+            } else {
+              throw e
```
ajimae, line 82: I think we need a test case here, to how the SDK handles this exception, hopefully in an integration test.

packages/sdk-client-v3/src/utils/validate.ts
ajimae: This was added in nodejs v15 [see docs](https://nodejs.org/api/globals.html#class-abortcontroller) nodejs version prior doesn't have this global object, hence the check we performed on line 24.

packages/sdk-client-v3/src/utils/validate.ts
```diff
@@ -19,11 +19,6 @@ export function validateHttpOptions(options: HttpMiddlewareOptions) {
...
-    throw new Error(
```
ajimae, line 24: Was there any reason for removing this?

packages/sdk-client-v3/tests/http.test/http-middleware.test.ts
```diff
@@ -52,23 +53,6 @@ describe('Http Middleware.', () => {
     ).toThrow()
   })
 
-  test('should throw if timeout is provided and `AbortController` is not.', () => {
```
ajimae, line 55: Also, any reason for removing this test case?

ajimae (changes requested):
Nice work @lojzatran 

I left a few comments in, please let me know if you need further clarification.

Cheers 🥂

packages/sdk-client-v3/src/utils/executor.ts
lojzatran: It's for consistency with v2: https://github.com/commercetools/commercetools-sdk-typescript/blob/68ad0e322b355cdee28e3114c95e60d0b9191df0/packages/sdk-client/src/sdk-middleware-http/http.ts#L91

packages/sdk-client-v3/src/types/types.d.ts
lojzatran: I added it here: https://github.com/commercetools/commercetools-docs/pull/4986/files#diff-4c9187ddfa33195642947e7508f80e1fced9d8e01127e80d1c4b2d9544f46524R18

packages/sdk-client-v3/src/types/types.d.ts
lojzatran: https://github.com/commercetools/commercetools-docs/pull/4986/commits/d763b02415c2bf6a4b13f27f38f3a12bd7b241fa

lojzatran:
> Nice work, I noticed that there is no integration tests to show (using **actual test data**) that this change actually works as intended?

I couldn't find a way to make CTP abort the connection, therefore there is no real test for this. If you have an idea how to do it, let me know.

packages/platform-sdk/test/integration-tests/sdk-v3/middlewares.test.ts
lojzatran: https://github.com/commercetools/commercetools-sdk-typescript/pull/732

packages/sdk-client-v3/src/utils/executor.ts
lojzatran: https://github.com/commercetools/commercetools-sdk-typescript/pull/714/commits/f435d6bd4df8dcc9b958d49fb2bb0199ed14d095

I added only a unit test. I can't find a reproduceable case when we would get this error with integration test. If you have a suggestion, let me know.

packages/sdk-client-v3/src/types/types.d.ts
ajimae: I know it was added there, but I am saying is that it should have also been added [here](https://github.com/commercetools/commercetools-docs/pull/4986/files#diff-d58e725f4184915964670f40fe874774b1dbdfbb8e1a06419683540c7b1481d1R255) in the difference between the v2 and v3 middleware.

packages/sdk-client-v3/src/utils/executor.ts
ajimae: Okay,  but since this is a new Major Version, I think some things have to change. I believe making this option true by default is better. Abort operation is usually network related error, so regardless of whether v2 is false by default I think it will make more sense to have this true here.

What do you think?

packages/sdk-client-v3/src/types/types.d.ts
lojzatran: Ah I see what you mean. I checked and it's not the right place to add it. The part where you pointed to is for migration from v2 to v3. Since this option was not available at all in v2, there is no need to migrate it.

packages/sdk-client-v3/src/utils/executor.ts
ajimae: Not really, I don't think these two are the same, considering where the both of them are situated. One the v2 is to retry a network error, in the finally error callback, inside that callback we are certain that it's a network error whilst in the v3 the check is done, in the executor (whose error might not specifically be a network error).

Also, the slightly different logic check here is telling, I can decide to have a retry for a 503 error (that is enableRetry will be set to true) but at the same time I don't want to enable any retryOnAbort (retryOnAbort will be false) at this combination, the customer won't get any retry. Is this correct?

I think we need to have a very good look at this check and all the possible truthy/falsey case for each combination of the individual variables like the `enableRetry`, `retryOnAbort` `abortController` etc, so we are sure the executor function behaves as intended.

packages/sdk-client-v3/tests/http.test/http-middleware.test.ts
ajimae: Not really, this can actually happen (in older node versions), this check was done for older nodejs versions that doesn't have the AbortController class built-in e.g nodejs v14 and older.

packages/sdk-client-v3/src/utils/executor.ts
ajimae: Sure, we can use the signal property to signal the AbortController to abort the request at any time using the abort timer. See my commend below.

ajimae:
> > Nice work, I noticed that there is no integration tests to show (using **actual test data**) that this change actually works as intended?
> 
> I couldn't find a way to make CTP abort the connection, therefore there is no real test for this. If you have an idea how to do it, let me know.

See this [here](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) for more on how you can combine the AbortSignal with an abort timer.

We can pass this as an option to the fetch function `{ signal: AbortSignal.timeout(5000) }` set the time to maybe `10` or even `0`. Then when the call is made, it immediately aborts the connection and since we have a check on the "AbortError" error name we can test this exception using an integration test.

ajimae:
@lojzatran I suggest we have a call to discuss some of these points in this PR, especially the test and the retry logic. 

Thanks 🙏🏽

ajimae:
> > > Nice work, I noticed that there is no integration tests to show (using **actual test data**) that this change actually works as intended?
> > 
> > 
> > I couldn't find a way to make CTP abort the connection, therefore there is no real test for this. If you have an idea how to do it, let me know.
> 
> See this [here](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) for more on how you can combine the AbortSignal with an abort timer.
> 
> We can pass this as an option to the fetch function `{ signal: AbortSignal.timeout(5000) }` set the time to maybe `10` or even `0`. Then when the call is made, it immediately aborts the connection and since we have a check on the "AbortError" error name we can test this exception using an integration test.

Hi @lojzatran See the snippet below, I was able to abort a requests successfully in the integration test. Do you think this will help us properly test the entire thing rather than testing just an isolated function?

```ts
describe('AbortController', () => {
  it('should abort request', () => {
    const client = new ClientBuilderV3()
      .withProjectKey(projectKey)
      .withClientCredentialsFlow(authMiddlewareOptionsV3)
      .withHttpMiddleware({
        ...httpMiddlewareOptionsV3,
        httpClient: function (url: string, ...args: any[]) {
          return fetch(url, { ...args, signal: AbortSignal.timeout(10) });
        }
      }).build();

    const api = createApiBuilderFromCtpClient(client);
    api.withProjectKey({ projectKey })
      .get()
      .execute()
      .then(console.log)
      .catch(e => {
        expect(e.name).toBe(...) // our custom error handler converted this into an `HttpError`
        console.log(e.name, '<>', e.message, '>>>>>>>>>>>>>>>>>>>>>>>>>')
        console.log(e)
      })
  })
})
```

Next we can then try to see if we can retry on this successful aborted request. Please let me know what you think if this is doable.

Thanks

lojzatran:
> > > > Nice work, I noticed that there is no integration tests to show (using **actual test data**) that this change actually works as intended?
> > > 
> > > 
> > > I couldn't find a way to make CTP abort the connection, therefore there is no real test for this. If you have an idea how to do it, let me know.
> > 
> > 
> > See this [here](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) for more on how you can combine the AbortSignal with an abort timer.
> > We can pass this as an option to the fetch function `{ signal: AbortSignal.timeout(5000) }` set the time to maybe `10` or even `0`. Then when the call is made, it immediately aborts the connection and since we have a check on the "AbortError" error name we can test this exception using an integration test.
> 
> Hi @lojzatran See the snippet below, I was able to abort a requests successfully in the integration test. Do you think this will help us properly test the entire thing rather than testing just an isolated function?
> 
> ```ts
> describe('AbortController', () => {
>   it('should abort request', () => {
>     const client = new ClientBuilderV3()
>       .withProjectKey(projectKey)
>       .withClientCredentialsFlow(authMiddlewareOptionsV3)
>       .withHttpMiddleware({
>         ...httpMiddlewareOptionsV3,
>         httpClient: function (url: string, ...args: any[]) {
>           return fetch(url, { ...args, signal: AbortSignal.timeout(10) });
>         }
>       }).build();
> 
>     const api = createApiBuilderFromCtpClient(client);
>     api.withProjectKey({ projectKey })
>       .get()
>       .execute()
>       .then(console.log)
>       .catch(e => {
>         expect(e.name).toBe(...) // our custom error handler converted this into an `HttpError`
>         console.log(e.name, '<>', e.message, '>>>>>>>>>>>>>>>>>>>>>>>>>')
>         console.log(e)
>       })
>   })
> })
> ```
> 
> Next we can then try to see if we can retry on this successful aborted request. Please let me know what you think if this is doable.
> 
> Thanks

Thanks for the hint. I added a test: https://github.com/commercetools/commercetools-sdk-typescript/pull/714/commits/d7ce9c7b403f104cc5ad08dc90b7093f712ff034

packages/sdk-client-v3/src/utils/executor.ts
lojzatran: https://github.com/commercetools/commercetools-sdk-typescript/pull/714/commits/f6b22c85647537d07299bea84be926ab116edacb

ajimae (approved):
🚀

packages/sdk-client-v3/src/utils/executor.ts
lojzatran: https://github.com/commercetools/commercetools-sdk-typescript/pull/714/commits/3c40d1de696e972de14c27a4dfc2e00b7af00c7c, https://github.com/commercetools/commercetools-docs/pull/4986/commits/a981b6b9b3bb15d679593f377fad85abea1fb86a

PR #720, opened 2024-05-27
Add Custom Metrics for Newrelic and Datadog

ajimae:
### Summary
Add agent for sending custom metrics to newrelic and datadog

### Completed Tasks
- [x] add service agents for individual provider
- [x] add usaage examples on how to export custom metrics

### Screenshots
Screenshot 2024-05-24 at 00 40 47

Screenshot 2024-05-24 at 00 41 15


lojzatran:
Please add a short README on how to run the examples as well. Sth like we did here: https://github.com/commercetools/commercetools-sdk-java-v2/tree/main/examples/spring-datadog

examples/datadog-express-apm/host-agent/utils/response.js
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    if (data) {
+      responseBody.data = data
+    }
+
+    agent
+      .init()
+      .dogstatsd.increment(
+        `Commercetools/Client/Response/Success/${statusCode}`,
```
lojzatran, line 33: Please update the names to the convention of datadog: https://docs.datadoghq.com/developers/guide/what-best-practices-are-recommended-for-naming-metrics-and-tags/

examples/datadog-express-apm/host-agent/utils/response.js
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    if (data) {
+      responseBody.data = data
+    }
+
+    agent
+      .init()
+      .dogstatsd.increment(
+        `Commercetools_Client_Response_Success_${statusCode}`,
```
lojzatran, line 33: I think it is better to have one metric with the name `Commercetools_Client_Response_Success` than multiple metrics with names like `Commercetools_Client_Response_Success_200`, `Commercetools_Client_Response_Success_201`, `Commercetools_Client_Response_Success_202` etc. The response code could be put as a tag. We also did it like this in java: https://github.com/commercetools/commercetools-sdk-java-v2/blob/main/commercetools/commercetools-monitoring-datadog/src/main/java/com/commercetools/monitoring/datadog/DatadogUtils.java#L43

The similar idea is applied also for error counters.

examples/newrelic-express-apm/src/utils/response.js
```diff
@@ -52,6 +58,11 @@ class ResponseHandler {
       responseBody.data = data
     }
 
+    agent.recordMetric(
+      `Commercetools/Client/Response/Error/${statusCode}`,
+      statusCode
```
lojzatran, line 63: Should this be a counter metric instead? Same for success metric above.

lojzatran (changes requested):
@ajimae good job 🍻 I see some small stuff missing:
1) Please add also Duration metric: https://docs.commercetools.com/sdk/observability/datadog
2) Are you going to create a separate PR for OpenTelemtry?

examples/datadog-express-apm/host-agent/utils/response.js
ajimae: Not really, that is not how it works, it will aggregate all the metrics with the same name and increment the count or average it, there are no sign of multiple occurrence of the name `Commercetools_Client_Response_Success`

You can look at the screenshot for reference.

examples/newrelic-express-apm/src/utils/response.js
ajimae: No, it shouldn't, take a look at [this docs](https://docs.newrelic.com/docs/apm/agents/nodejs-agent/extend-your-instrumentation/nodejs-custom-metrics/#example) maybe you will understand why it should be `recordMetric` method. Also, It is important that you know that there are different method implementation for the agents for individual languages.

examples/datadog-express-apm/host-agent/README.md
```diff
@@ -0,0 +1,62 @@
...
+- For Datadog setup, follow the instructions stated in our [official documentation website](https://docs.commercetools.com/sdk/observability/datadog#typescript-sdk) to properly install and set up the Datadog agent.
+
+## Installation
+
+1. Clone/Download the example folder.
+2. Navigate to the path `datadog-express-apm/host-agent/`.
+3. Create a `.env` file in this path, add and update the content of `.env.sample`, the `DD_API_KEY` is required
+4. Run `yarn install` (`npm` can also be used) to install all dependencies
+
+```bash
```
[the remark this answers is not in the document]
lojzatran: Sorry probably I was not clear. If you read this doc and go step by step in the Installation part, after you read 4., you suddenly see this bash part: https://github.com/commercetools/commercetools-sdk-typescript/blob/88cbf0137d8be8a17f1caecbefe532337e9e33e0/examples/datadog-express-apm/host-agent/README.md. Does it belong to point 4? I guess it's point 3. But why would you check it here when you have the same content in `.env.sample` that you need to open anyway and change it there?

examples/datadog-express-apm/host-agent/utils/response.js
lojzatran: But the screenshots are from NewRelic, not datadog. Can you post screenshots from Datadog as well? The difference between datadog and newrelic custom metrics is that in datadog you can add tags (https://docs.datadoghq.com/metrics/custom_metrics/). On the other hand, NewRelic has only time slice metrics, therefore naming convention there is different: https://docs.newrelic.com/docs/apm/agents/manage-apm-agents/agent-data/collect-custom-metrics/

ajimae: I don't know if you can see it, but I can, and there's no duplicate there. If you are in doubt please clone the example repo, run it and show me the duplicate.

Screenshot 2024-05-27 at 11 09 02

jenschude: The image is blurry. But Lam is correct here. The status code should be part of the metrics attributes. This way you can have a total view of all requests and will be able to filter/group the metrics by status code. If it's part of the name it's harder to combine them in one view.

Should be something like this:

```
.dogstatsd.increment(
        `Commercetools_Client_Response`,
        1,
        { env: 'dev', status_code': statusCode, http_method: req.method, success: true  }
      )

jenschude (approved):

PR #796, opened 2024-09-02
fix(tests): check before create new product type

lojzatran:
There was an issue that more than 1000 product types were created. This is the limit of CTP: https://docs.commercetools.com/api/limits#product-types. So, we need to ensure that the product type is created only if it does not exist. This PR checks if the product type exists before creating a new one.


ajimae:
Hi @lojzatran, can we get a description of what this PR or maybe what it is trying to achieve? Thanks

lojzatran:
> Hi @lojzatran, can we get a description of what this PR or maybe what it is trying to achieve? Thanks

Hi @ajimae, I added the description.

barbara79 (approved):
it makes totally sense to me. 
Thanks 💯

PR #815, opened 2024-10-04
feat(cleanup): add bulk resource deletion script for integration tests (DEVX-433)

evansinho:
This PR introduces a script to handle the bulk deletion of various resources, ensuring the system stays clean and within resource limits. Key features include:

- Unpublishing and deleting all products.
- Safe deletion of resources, skipping any referenced by others (e.g., tax categories, customer groups, stores).
- Improved error handling and logging for easier tracking.
- Resolved versioning conflicts during resource updates to prevent concurrency issues.
- This will help maintain a cleaner environment and avoid hitting commercetools resource limits.

### DIFF
--- a/packages/platform-sdk/test/integration-tests/CleanEverythingIntegrationTest.ts
+++ b/packages/platform-sdk/test/integration-tests/CleanEverythingIntegrationTest.ts
@@ -1,0 +1,242 @@
+import { Product } from '../../src'
+import { apiRoot } from './test-utils'
+
+// Generic function to handle deletion of resources
+const deleteResource = async (resourceType, resourceId, version) => {
+  try {
+    await apiRoot[resourceType]()
+      .withId({ ID: resourceId })
+      .delete({ queryArgs: { version } })
+      .execute()
+    console.log(`Deleted ${resourceType.slice(0, -1)} with ID: ${resourceId}`)
+  } catch (error) {
+    console.error(
+      `Error deleting ${resourceType.slice(0, -1)} with ID: ${resourceId}:`,
+      error
+    )
+  }
+}
+
+// Function to unpublish and delete a product
+const unpublishAndDeleteProduct = async (product: Product) => {
+  if (product.masterData.published) {
+    console.log(`Unpublishing product: ${product.id}`)
+    await apiRoot
+      .products()
+      .withId({ ID: product.id })
+      .post({
+        body: {
+          version: product.version,
+          actions: [{ action: 'unpublish' }],
+        },
+      })
+      .execute()
+    console.log(`Product ${product.id} unpublished.`)
+  }
+
+  // Increment version after unpublish
+  await deleteResource(
+    'products',
+    product.id,
+    product.version + (product.masterData.published ? 1 : 0)
+  )
+}
+
+// Function to delete all products for a specific product type
+const deleteProductsForProductType = async (productTypeId: string) => {
+  const products = await apiRoot
+    .products()
+    .get({ queryArgs: { where: `productType(id="${productTypeId}")` } })
+    .execute()
+
+  console.log(
+    `Found ${products.body.results.length} products for product type ${productTypeId}...`
+  )
+  return Promise.all(
+    products.body.results.map((product: Product) =>
+      unpublishAndDeleteProduct(product)
+    )
+  )
+}
+
+// Function to delete all products in the system
+const deleteAllProducts = async () => {
+  const products = await apiRoot.products().get().execute()
+  console.log(`Found ${products.body.results.length} products to delete...`)
+
+  return Promise.all(
+    products.body.results.map(async (product: Product) => {
+      await unpublishAndDeleteProduct(product)
+    })
+  )
+}
+
+// Generic function to check if a resource is referenced by another resource and delete it if not
+const safeDeleteResource = async (
+  resourceType,
+  resourceId,
+  version,
+  referenceCheckFn
+) => {
+  const isReferenced = await referenceCheckFn()
+  if (isReferenced) {
+    console.log(
+      `${resourceType.slice(0, -1)} with ID: ${resourceId} is referenced. Skipping deletion.`
+    )
+    return
+  }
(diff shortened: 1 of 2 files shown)


packages/platform-sdk/test/integration-tests/CleanEverythingIntegrationTest.ts
```diff
@@ -0,0 +1,242 @@
...
+    // Product Types and associated products
+    const productTypes = await apiRoot.productTypes().get().execute()
+    await Promise.all(
+      productTypes.body.results.map(async (productType) => {
+        await deleteProductsForProductType(productType.id)
+      })
+    )
+    console.log(`Deleted ${productTypes.body.results.length} Product Types`)
+
+    // Delete all standalone products (not tied to a product type)
```
cneijenhuis, line 103: Sorry, can you clarify this point - I thought products need always to be tied to a product type? https://docs.commercetools.com/api/projects/products#product

packages/platform-sdk/test/integration-tests/CleanEverythingIntegrationTest.ts
```diff
@@ -0,0 +1,242 @@
...
+    // Tax Categories
+    const taxCategories = await apiRoot.taxCategories().get().execute()
+    await Promise.all(
+      taxCategories.body.results.map(async (taxCategory) => {
+        await safeDeleteResource(
+          'taxCategories',
+          taxCategory.id,
+          taxCategory.version,
+          async () => {
+            const productsReferencingTaxCategory = await apiRoot
```
cneijenhuis, line 152: I'm not following this for several reasons:
- At this point, we should have already deleted all products - why should we double-check here?
- Why are you trying to duplicate the "safe deletion" logic here? If it isn't safe to delete, the API should refuse to delete (and you should react to the error). E.g. for tax categories, [shipping methods also reference them](https://docs.commercetools.com/api/projects/shippingMethods#shippingmethod), and the API will refuse the request. Isn't it better to react to that error?

packages/platform-sdk/test/integration-tests/CleanEverythingIntegrationTest.ts
evansinho: @cneijenhuis Thanks for pointing that out! Yes, you're right, products are always tied to a product type. However, when deleting a product, we don't need to reference the product type explicitly in the deletion process. In this case, I'm fetching all the products and deleting them directly. The link to the product type is part of the product’s metadata, but it doesn’t need to be called out specifically for the deletion to happen. I focus on unpublishing and then deleting the product itself. 

Let me know if that makes sense!

packages/platform-sdk/test/integration-tests/CleanEverythingIntegrationTest.ts
evansinho: - You're right that by the time we get to tax categories, we should have already deleted all products, and ideally, there wouldn't be any remaining references. The reason I added that check was to be extra cautious in case anything slipped through or there were edge cases where products might still reference a tax category. It’s more of a precaution to prevent unexpected errors during the cleanup.
- As for duplicating the "safe deletion" logic, I agree...it’s better to rely on the API’s error handling when trying to delete something that’s still referenced, like in the case of shipping methods referencing tax categories. I can adjust the code to simply handle those errors directly rather than pre-checking, which will make the process cleaner and more efficient.

ajimae:
Nice job 👍🏽

I was looking at the jira ticket and I could see that other methods were also hinted. I think that aside this standalone script solution, we should also have in addition test `hooks` for relevant tests suites such as a `beforeAll` and `afterAll` hooks that does this just as we have [here](https://github.com/commercetools/commercetools-sdk-typescript/blob/511ba9acbe66797577790dce35d3c9aa12a0cd93/packages/platform-sdk/test/integration-tests/sdk-v3/middlewares.test.ts#L36-L55)

We created the resource we needed for the test and immediately delete the resource(s) after the tests finished running. This way we are always sure of a clean slate anytime we are running the integration tests.

What do you think?

evansinho:
@ajimae  Thanks for the feedback! 👍🏽 I completely agree. Having the beforeAll and afterAll hooks for the test suites to handle resource creation and deletion will ensure a clean slate for each test run, making the process more robust. I'll create a new branch and start implementing this approach alongside the standalone script solution.

packages/platform-sdk/test/integration-tests/CleanEverythingIntegrationTest.ts
```diff
@@ -0,0 +1,242 @@
...
+const deleteProductsForProductType = async (productTypeId: string) => {
+  const products = await apiRoot
+    .products()
+    .get({ queryArgs: { where: `productType(id="${productTypeId}")` } })
+    .execute()
+
+  console.log(
+    `Found ${products.body.results.length} products for product type ${productTypeId}...`
+  )
+  return Promise.all(
```
lojzatran, line 55: It's not recommended to use `Promise.all` with requests to commercetools. Rather use [Promise.map](http://bluebirdjs.com/docs/api/promise.map.html) and limit the concurrency to some lower number like 20. The reason is CTP cannot handle well the deletion and deletion usually triggers some backend work as well. I know this because I did `Promise.all` as well and caused the whole platform outage in the past 😅

packages/platform-sdk/test/integration-tests/CleanEverythingIntegrationTest.ts
lojzatran: @evansinho I would recommend to delete straightaway and not checking. 
Reasons:
- At this point products are all delete. If not, then it should fail and you should investigate why. It could be an error in the script and then we have to fix it. If it's a platform error, then we should report it.
- Checking this requires extra request/response. This script could be used as a reference for customers, so we should avoid doing unnecessary actions.

Please check also for other resources that use `safeDeleteResource()`.

lojzatran:
I‘m closing this PR since it's outdated and @evansinho won't work on this anymore.

PR #841, opened 2024-11-04
[Chore][DEVX-476] Improve README file

ajimae:
### Summary
Improve SDK `README.md` file, remove and/or update outdated code and documentations

### Completed Tasks
- [x] Remove outdated docs
- [x] Add more use case Improve readme file

### Jira Ticket
[DEVX-476](https://commercetools.atlassian.net/browse/DEVX-476)

[DEVX-476]: https://commercetools.atlassian.net/browse/DEVX-476?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/sdk-client-v3/README.md
```diff
@@ -150,11 +154,13 @@ const client = new ClientBuilder()
   .build()
 ```
 
-> [!WARNING]
-> Do not add the built-in middlewares using `withMiddleware` method. Adding by this method does not respect the ordering of the middlewares and could lead to unexpected behavior.
+<!-- > [!WARNING]
```
lojzatran, line 157: Is this not true anymore? I still see the ordering is taken into account in this method: https://github.com/commercetools/commercetools-sdk-typescript/blob/9c4f95a64c5469ad6c686b90aae8908bd440433e/packages/sdk-client-v3/src/client/builder.ts#L266.

Will it work now if I add middlewares in any order? Or is this ordering somewhere written?

ajimae: I believe I [mentioned](https://github.com/commercetools/commercetools-sdk-typescript/pull/841/files#diff-9cac8b52494830bd5b1a052931412e3d854b4e7ee558d49eb781fbeacd1dadb6R207) it that when using only the `withMiddleware()` method the order matters, hence the need to use only the `withMiddleware()` method to add all middleware to the call chain.

If you see in the example, all the middleware were added using the `withMiddleware()` method whilst the second example shows what would happen if this middleware methods are mixed.

lojzatran: > The `withMiddleware` method can be used to add middleware functions (both built-in and custom middleware) in an ordered fashion.

From this sentence I understand that `withMiddleware` adds the middlewares in some order, but it is not clear what is the order or who creates this order.

I think the main point here is that the middlewares would not work correctly if added using `withMiddleware()` in random order. There is no section about how the built-in middlewares should be ordered to work correctly.

ajimae: Oh I can see what you mean now. I will create a task to change this behaviour in the SDK. We have to find a way to allow the customers to add custom middleware at any point in the execution chain. We always had this feature pre SDK v2.

lojzatran (approved):

PR #864, opened 2024-12-03
[Fix](DEVX-364): Fixing support ticket ST-29803

ajimae:
### Summary
Fix issues with `sdk-client-v3` returning`200 - OK` status with an `undefined` body.

### Completed Task
- [x] throw `un-parsable` response body as error

### Jira Ticket
[#DEVX-364](https://commercetools.atlassian.net/jira/software/c/projects/DEVX/boards/386?selectedIssue=DEVX-364)

### Support Ticket
[#ST-29803](https://commercetools.atlassian.net/browse/SUPPORT-29803?focusedCommentId=858951)


lojzatran (approved):
LGTM 👍 

But how this fixes the issue from the support ticket? Customer is describing that the body is empty on 200 status code, and here is about the error body?

ajimae:
> LGTM 👍
> 
> But how this fixes the issue from the support ticket? Customer is describing that the body is empty on 200 status code, and here is about the error body?

This is the fix on this line, so instead of reassigning the result to data, we just throw the error. This error arises when the request is aborted too soon, as you can see there is no handler for it.

PR #877, opened 2024-12-11
fix(DEVX-363): remove node 16 support

lojzatran:
JIRA: https://commercetools.atlassian.net/browse/DEVX-363

1. I removed fetch package as it's now supported in nodejs and in major browsers as well: https://caniuse.com/?search=fetch. However, there is [an issue that current nock version cannot mock the routes anymore](https://github.com/nock/nock/issues/2397). After some searching I found out that they're adding mocking for fetch in the newest beta of nock, so I did the upgrade. https://github.com/nock/nock/releases/tag/v14.0.0-beta.1

2. There is a change in the error message when aborting connection between nodejs versions. This makes our code not working properly: https://github.com/nodejs/undici/issues/2590. After fixing it, I was thinking if we should actually run integration tests for all versions because this change is an example that there are changes in versions that could affect us.

Q: should we run integration tests for all supported nodejs versions?


packages/sdk-client-v3/src/utils/executor.ts
```diff
@@ -106,7 +105,11 @@ export default async function executor(request: HttpClientConfig) {
               return { _response, shouldRetry: true }
             }
           } catch (e) {
-            if (e.name.includes('AbortError') && retryWhenAborted) {
+            if (
+              (e.name.includes('AbortError') ||
```
ajimae, line 109: Is it possible to add a comment here to explain why we needed this two different error names and which error name belongs to which nodejs version ?

packages/sdk-client/test/auth.test/base-auth-flow.test.ts
```diff
@@ -455,6 +455,7 @@ describe('Base Auth Flow', () => {
       // await promise failure which should set `requestState` to `false`
       await startCreateBaseMiddleware()
     } catch (error) {
+      expect(requestCount).toBe(1) // Make sure that nock runned the mocked request
```
ajimae, line 458: Why was this moved to the catch blocked?

packages/sdk-client/test/auth.test/base-auth-flow.test.ts
```diff
@@ -414,22 +413,24 @@ describe('Base Auth Flow', () => {
...
     nock(middlewareOptions.host)
       .persist() // <-- use the same interceptor for all requests
-      .log(() => {
-        requestCount += 1
-      }) // keep track of the request count
       .filteringRequestBody(/.*/, '*')
-      .post('/oauth/token', '*')
+      .post('/oauth/token', () => {
+        requestCount += 1
+        return true
```
ajimae, line 419: Why do we have this here. I think you should remove this `true` so we can determine when the post errors out.

lojzatran: This is to match anything in the body of `post`. It has the same meaning as `*`. I just make it a function and the function must return boolean if the body matches or not. It has nothing to do with errors.

packages/sdk-client/test/auth.test/base-auth-flow.test.ts
lojzatran: 1. It actually makes more sense. This test is named `when createAuthMiddlewareBase fails`, so the verification should be in catch if possible.
2. Before we were using `log` to increase the requestCount, but this method is now removed. I have to move this requestCount increasing to `post`. However, this increasing now works at a different moment than with `log`, so this expect cannot be at its original position anymore.

packages/sdk-client/test/http.test/http.test.ts
```diff
@@ -1351,9 +1341,7 @@ describe('Http', () => {
             expect(res.error.headers).toBeUndefined()
             expect(res.error.originalRequest).toBeDefined()
             expect(res.error.retryCount).toBe(2)
-            expect(res.error.message).toBe(
-              `request to ${testHost}/foo/bar failed, reason: Connection timeout`
-            )
+            expect(res.error.message).toBe('Connection timeout')
```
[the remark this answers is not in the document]
lojzatran: This message does not come from our code or at least I cannot find it anywhere. I guess there could be a change in nodejs, but I didn't look exactly where this change has been done.

packages/sdk-client-v3/src/utils/executor.ts
lojzatran: https://github.com/commercetools/commercetools-sdk-typescript/pull/877/commits/fa89ac7e5e2d1b8f3cc4d1a1d16d029607439340

PR #884, opened 2024-12-16
[Fix][DEVX-514] Fix v3 AbortController Retry Reinitialization

ajimae:
### Summary
Fix issue with abortController always reusing the inital timedout controller

### Completed Tasks
- [x] fix issue with abortController not reinitialized for each request

### Support Ticket
[#30038](https://commercetools.atlassian.net/browse/SUPPORT-30038?focusedCommentId=865938)

### TODO
- [ ] Include tests

### DIFF
--- a/.changeset/grumpy-ways-drum.md
+++ b/.changeset/grumpy-ways-drum.md
@@ -1,0 +1,5 @@
+---
+"@commercetools/ts-client": patch
+---
+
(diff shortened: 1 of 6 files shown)


lojzatran:
I did a manual test and the problem is still there. As the customer pointed out in the ticket, `AbortController` needs to be reinitialized and reassigned as well, along with `signal`. I will make an additional fix now.

lojzatran (approved):

PR #887, opened 2024-12-20
fix(SUPPORT-30038): clone deep the request object

lojzatran:
Fixes: https://commercetools.atlassian.net/browse/SUPPORT-30038?focusedCommentId=865938

The issue was when masking the auth data, we also update the original Authorization header in the request. Then when we retry the request, the request does not have the original Authorization header anymore.
To fix it, I make a deep clone of the request object. I tried to use `structuredClone()` method built in nodejs, but this throws error when you have a function in a cloned object (and we have `resolve` and `reject` inside `request`). So I had to use lodash clone function instead.

### DIFF
--- a/packages/sdk-client-v3/src/utils/maskAuthData.ts
+++ b/packages/sdk-client-v3/src/utils/maskAuthData.ts
@@ -1,8 +1,7 @@
 import { MiddlewareRequest } from '../types/types'
-import cloneDeep from 'lodash.clonedeep'
 
 export default function maskAuthData(request: MiddlewareRequest) {
-  const _request = cloneDeep(request)
+  const _request = JSON.parse(JSON.stringify(request))
   if (_request?.headers) {
     if (_request.headers.Authorization) {
       _request.headers['Authorization'] = 'Bearer ********'
--- a/packages/sdk-client-v3/src/utils/maskAuthData.ts
+++ b/packages/sdk-client-v3/src/utils/maskAuthData.ts
@@ -1,7 +1,8 @@
 import { MiddlewareRequest } from '../types/types'
+import cloneDeep from 'lodash.clonedeep'
 
 export default function maskAuthData(request: MiddlewareRequest) {
-  const _request = Object.assign({}, request)
+  const _request = cloneDeep(request)
   if (_request?.headers) {
     if (_request.headers.Authorization) {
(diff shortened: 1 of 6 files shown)


packages/sdk-client-v3/src/utils/maskAuthData.ts
```diff
@@ -1,7 +1,8 @@
 import { MiddlewareRequest } from '../types/types'
+import cloneDeep from 'lodash.clonedeep'
 
 export default function maskAuthData(request: MiddlewareRequest) {
-  const _request = Object.assign({}, request)
+  const _request = cloneDeep(request)
```
ajimae, line 5: Instead of adding a `library` just to deep clone, I think a simple `JSON.parse(JSON.stringify(request))` will work.

ajimae: We can try as much as we can to avoid using third party `dependencies`.

This I believe the below snippet should also be sufficient, since `JSON.parse(JSON.stringify())` have issue with `functions` and `undefined` values, even though we don't have functions in our request object it is always good to be on the safer side.

We can run the the tests case again it as well.


```ts
import { MiddlewareRequest } from '../types/types'

export function deepCopy(request: MiddlewareRequest) {
  let map: MiddlewareRequest | Array<any>, index: string | number

  /** these checks can be ignored, since the function is not exposed */
  if (!request) return request
  if (typeof request !== 'object') return request

  if ('[object Array]' === Object.prototype.toString.apply(request)) {
    map = []
    for (index = 0; index < request.length; index++) {
      map[index] = deepCopy(request[index])
    }

    return map
  }

  map = Object.create(null)
  for (index in request) {
    if (request.hasOwnProperty(index)) {
      map[index] = deepCopy(request[index])
    }
  }

  return map
}
```

lojzatran: You are right, there‘s no need for 3rd party lib: https://github.com/commercetools/commercetools-sdk-typescript/pull/887/commits/320fa7a2facd8d9b9e9026171fb91cbb716f917a. I think we could just leave the function simple since the request structure is defined in TS and should work fine.

ajimae (approved):
LGTM 🚀

PR #902, opened 2025-01-12
[Chore][DEVX-522] Remove Buffer Module

### LINKED ISSUE #886: `require` call in ESM build of `ts-client`
We're using SvelteKit (which uses Vite) with the `@commercetools/ts-client` package. When we run the production version of the website it throws the following error when trying to import the package:

```
ReferenceError: require is not defined in ES module scope, you can use import instead
This file is being treated as an ES module because it has a '.js' file extension and '/app/package.json' contains "type": "module". To treat it as a CommonJS script, rename it to use the '.cjs' file extension.
```

After digging into this a bit the issue appears to be caused by the fact that
the ESM build of `ts-client` (`commercetools-ts-client.esm.js`) has a `require` call in it. E.g. this is how it appears in my node_modules directory:

image

The `buffer` package appears to allow using the node buffer module in the browser. We're using the package in node (server-side in SvelteKit), so if I remove the `require` call everything seems to work fine.

We've managed to work around this by using `patch-package`, but it doesn't seem correct the ESM file has a `require` call in it instead of an import statement?

The require call looks to have been added here: https://github.com/commercetools/commercetools-sdk-typescript/pull/835

Thanks!

ajimae:
### Summary
Remove all buffer module instance

### Completed task
- [x] remove all instance of the buffer module
- [x] test to ensure compatibility with client side applications

### Fixes
#886


packages/sdk-client-v3/src/utils/byteLength.ts
```diff
@@ -1,8 +1,10 @@
-var Buffer = require('buffer/').Buffer
-
 export default function byteLength<T>(body: T): string {
-  if (body && (typeof body === 'string' || body instanceof Uint8Array)) {
-    return Buffer.byteLength(body).toString()
+  if (body && typeof body === 'string') {
+    return body.normalize('NFD').length.toString()
+  }
+
+  if (body && body instanceof Uint8Array) {
```
lojzatran, line 6: I'm just curious, in which case we can get body as Uint8Array?

lojzatran (approved):
👍

packages/sdk-client-v3/src/utils/byteLength.ts
ajimae: For images, we have an image endpoint that will fail is this check if not done.

```ts
// Something like this...
...
  return apiRoot
    .withProjectKey({ projectKey })
    .products()
    .withId({ ID: '' })
    .images()
    .post({
      headers: {
        "Content-Type": "image/jpeg"
      },
      body: fileFromSync(path.resolve(__dirname, 'image.jpeg'), 'image/jpeg'),
      queryArgs: {
        filename: 'DKNY.jpeg'
      }
    })
    .execute();
}
...
```

barbara79 (approved):

PR #914, opened 2025-01-20
[Fix][DEVX-531] Issue with Response Serialization

ajimae:
### Summary
Properly serialize raw response JSON object for `HEAD` requests

### Completed Tasks
- [x] serialize the JSON object from the response when body is empty
- [x] include test to validate fix

### DIFF
--- a/packages/sdk-client-v3/src/utils/executor.ts
+++ b/packages/sdk-client-v3/src/utils/executor.ts
@@ -157,7 +157,7 @@ export default async function executor(request: HttpClientConfig) {
         if (response.text && typeof response.text == 'function') {
           result =
             (await response.text()) ||
-            response[Object.getOwnPropertySymbols(response)[1]]
+            JSON.stringify(response[Object.getOwnPropertySymbols(response)[1]])
           data = JSON.parse(result)
         } else {
           // axios response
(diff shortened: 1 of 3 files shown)


packages/sdk-client-v3/src/utils/executor.ts
```diff
@@ -157,7 +157,7 @@ export default async function executor(request: HttpClientConfig) {
         if (response.text && typeof response.text == 'function') {
           result =
             (await response.text()) ||
-            response[Object.getOwnPropertySymbols(response)[1]]
+            JSON.stringify(response[Object.getOwnPropertySymbols(response)[1]])
```
lojzatran, line 160: Why do you need stringify if the body for HEAD method is empty?

lojzatran (approved):

packages/sdk-client-v3/src/utils/executor.ts
ajimae: So the response doesn't just contain only the body, the headers, statusText and statusCode are still in the present in the response object, so it still needs to be parsed, since the `response[Object.getOwnPropertySymbols(response)[1]]` doesn't return a string by default, we need to stringify the returned object so that the `JSON.parse()` that follows won't throw an error.

Another way we could have done this is to check if it's a string before calling the JSON.parse() on the result, but I don't thing this is much of a case to introduce branching.

PR #919, opened 2025-01-27
DEVX-523 Fix createTelemetryMiddleware and migrate from client v2 to v3 datadog demo app

barbara79:
- createTelemetryMiddleware was mandatory asking for a NEW RELIC environment variable althought, the demo app tested is for Datadog.
It has been changed the logic.

- Linked to https://commercetools.atlassian.net/browse/DEVX-529 - since it's about the migration to v2 to v3 for DAtadog demo app


examples/datadog-express-apm/host-agent/src/sdk-v3.js
```diff
@@ -16,24 +18,27 @@ const authMiddlewareOptions = {
...
   host: 'https://api.europe-west1.gcp.commercetools.com',
   includeRequestInErrorResponse: false,
   includeOriginalRequest: true,
-  fetch,
+  httpClient: fetch,
 }
 
-// datadog options
-const agent = require('path').join(__dirname, '..', 'tracer.js')
+// newrelic options
```
lojzatran, line 31:
```suggestion
// datadog options
```

examples/datadog-express-apm/host-agent/src/sdk-v3.js
```diff
@@ -16,24 +18,27 @@ const authMiddlewareOptions = {
...
 const telemetryOptions = {
   createTelemetryMiddleware,
   userAgent: 'typescript-sdk-middleware-datadog',
-  tracer: require(agent).init(),
-  // apm: () => require('dd-trace').init()
+  tracer: () => tracer.init(),
+  apm: () => require('dd-trace').init(),
```
lojzatran, line 36: There is a file in this folder called `agent.js` which contains one line: `module.exports = require('dd-trace')`. Is the purpose to use this file only and not require `dd-trace` directly? I'm actually still not sure why we need this file. Can you check?

packages/ts-sdk-apm/src/apm.ts
```diff
@@ -28,18 +15,16 @@ export default function createTelemetryMiddleware(
   function trace() {
     // validate apm and tracer
     if (!(options?.apm && typeof options.apm == 'function')) {
-      options.apm = defaultOptions.apm
+      console.error('Apm is not a function.')
```
lojzatran, line 18: Why not throwing the error so that the init process fails?

packages/ts-sdk-apm/src/apm.ts
```diff
@@ -55,9 +40,14 @@ export default function createTelemetryMiddleware(
 
       // send `response_time` to APM platforms
       if (options?.customMetrics) {
-        options.customMetrics.newrelic && recordNewrelic(response_time)
-        options.customMetrics.datadog &&
-          recordDatadog(response_time, { env: 'dev' })
+        if (options.customMetrics.datadog) {
```
lojzatran, line 43: Is this option `customMetrics` documented? I couldn't find it in the docs: https://docs.commercetools.com/sdk/observability/datadog#include-the-datadog-middleware-in-the-typescript-sdk

Also, in the docs link above it says that we are adding automatically 3 metrics for datadog (`commercetools_client_response<Success>`, `commercetools_client_response<Error>`, `commercetools_client_request_total`). However, I could not find a place where this is done. I could see we are adding with slightly different letter case the metric `Commercetools_Client_Response_Total`, but I can't find the other 2 metrics. Can you point to the place in the SDK?

examples/datadog-express-apm/host-agent/README.md
```diff
@@ -8,6 +8,7 @@ Example to show how the Datadog APM can be used in the TypeScript SDK.
 - Your Project must have existing Products containing Variants, and at least one Customer.
 - If your Project is currently empty, you can install the [SUNRISE sample data](https://github.com/commercetools/commercetools-sunrise-data).
 - For Datadog setup, follow the instructions stated in our [official documentation website](https://docs.commercetools.com/sdk/observability/datadog#typescript-sdk) to properly install and set up the Datadog agent.
+  - Make sure that Node js has been installed as Integration to run this Demo application.
```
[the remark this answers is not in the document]
barbara79: There are Integrations in the Datadog UI that can be installed in this case the customer has to install the Nodejs integration.
I'm gonna add this [link](https://docs.datadoghq.com/integrations/) to be sure that it's more clear.
For your point of view it would be enough?

packages/ts-sdk-apm/src/apm.ts
barbara79: Thank you for noticing it!
I also didn't find them ;) and this code was already there. I am gonna add this to the [task](https://commercetools.atlassian.net/browse/DEVX-524)

barbara79:
> Good job 👍 I have some additional questions:
> 
> 1. Why didn't you upgrade the container-agent for datadog as well?
> 2. What should the user see if they run this demo example? Can you give a screenshot and names of the metrics, log or traces examples?

1. I didn't because since I modified some files in the package @commercetools/ts-sdk-apm, Docker takes the files from the npm released library, so I could change them but I cannot test it in the Docker environment.
2. This is what the customer could see.
Screenshot 2025-01-27 at 15 07 31
Screenshot 2025-01-27 at 15 07 52
I was thinking to write a wiki page in Confluence with those kind of information or see if I can add this 2 images in the readme file. WDYT?

examples/datadog-express-apm/host-agent/src/sdk-v3.js
barbara79: yes, you are right. I believe it was required since there was a default value in the ts-sdk-apm for apm/tracer. I'm gonna remove the agent.js file.

packages/ts-sdk-apm/src/apm.ts
barbara79: About the second question, it's not done in the SDK, but in the demo app itself:
please check the response.js file and in the app.js file.

barbara79: Yes, it makes sense, I am creating in parallel a branch in the docs repo to update these kind of information ;)

barbara79:
> Good job 👍 I have some additional questions:
> 
> 1. Why didn't you upgrade the container-agent for datadog as well?
> 2. What should the user see if they run this demo example? Can you give a screenshot and names of the metrics, log or traces examples?

This is the [wiki page created](https://commercetools.atlassian.net/wiki/spaces/CTCLIENTS/pages/1575813203/Datadog+for+TS+SDK). 
We could add other information there for our knowledge

packages/ts-sdk-apm/src/apm.ts
barbara79: Also because the official documentation is not everywhere updated with the new client

packages/ts-sdk-apm/src/apm.ts
```diff
@@ -28,18 +15,22 @@ export default function createTelemetryMiddleware(
...
+      )
+      options.apm = () => {}
     }
 
     if (!(options?.tracer && typeof options.tracer == 'function')) {
-      options.tracer = defaultOptions.tracer
+      console.warn(
+        'Tracer is missing or not a function. Proceeding without tracing.'
+      )
+      options.tracer = () => {}
```
barbara79, line 28: I added a warning so it want totally stop the application

examples/datadog-express-apm/host-agent/package.json
```diff
@@ -1,11 +1,11 @@
 {
-  "name": "datadog",
+  "name": "my-app",
```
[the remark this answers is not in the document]
barbara79: see [here](https://github.com/commercetools/commercetools-sdk-typescript/pull/919/files#diff-ec0cc66a2d4870dd66d2cd5f26bf8421cee2e6632690d973edf9f54ba73c58cd)

packages/ts-sdk-apm/src/apm.ts
ajimae: Please can we remove any console statement from code, it is unnecessary. We cannot be logging statements in the customer's applications.

I believe on the review call we had last week on this, I explain to you what to do here. Just for clarification:

- Remove the console statement(s)
- Default these functions (tracer and apm) using an empty functions, so even if there is no attached collector the function can just fall through.

packages/ts-sdk-apm/test/apm.test/apm.test.ts
```diff
@@ -41,19 +41,6 @@ describe('apm', () => {
 
       await telemetryMiddleware(next)(request)
     })
-
-    test('should use default apm and tracing configurations', async () => {
```
ajimae, line 45: If I may ask, why was this test removed?

examples/datadog-express-apm/host-agent/utils/response.js
```diff
@@ -30,7 +30,7 @@ class ResponseHandler {
     agent.init().dogstatsd.increment(`Commercetools_Client_Response`, 1, {
       env: 'dev',
       status_code: statusCode,
-      http_method: req.method,
```
ajimae, line 33: Why was this remove?

examples/datadog-express-apm/host-agent/src/sdk-v3.js
```diff
@@ -16,23 +18,25 @@ const authMiddlewareOptions = {
...
 
 // datadog options
-const agent = require('path').join(__dirname, '..', 'tracer.js')
 const telemetryOptions = {
   createTelemetryMiddleware,
   userAgent: 'typescript-sdk-middleware-datadog',
-  tracer: require(agent).init(),
-  // apm: () => require('dd-trace').init()
+  tracer: () => tracer.init(),
+  apm: () => require('dd-trace').init(),
```
ajimae, line 36: We can remove the `apm: () => require('dd-trace').init()` update the `tracer: () => require('dd-trace').init()` also, since you introduced the lazy module import, can you also update all the `require` statements to use this lazy import.

```suggestion
  tracer: async () => await import('dd-trace').init()
```
Also, I didn't test this, but I believe it should work. Please verify that it actually works.

ajimae (commented):
Nice work @barbara79 

I left a few comments, please take a look at them.

Thank you

examples/datadog-express-apm/host-agent/src/sdk-v3.js
barbara79: So about your suggestion, I believe it's better to have it as it is so the customers can have the chance to modify tracer.js since this is just and example for them. What do you think?

About `can you also update all the require statements to use this lazy import.`, what do you mean? could you give me an example?

examples/datadog-express-apm/host-agent/utils/response.js
barbara79: because it was giving me the error on the `req` variable that is not passed

examples/datadog-express-apm/host-agent/utils/response.js
```diff
@@ -63,7 +63,7 @@ class ResponseHandler {
...
```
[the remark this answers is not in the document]
barbara79: because it was giving me the error on the req variable that is not passed and declared

packages/ts-sdk-apm/src/apm.ts
barbara79: so customers cannot see any traces, right?

ajimae: I mean you should default these to a default trace collector and remove the `console.warn` so even if the customer doesn't pass any tracer function as an option, we should still be able to record using our default tracer function, if you don't want to default it then fine, just go ahead and remove the console.warn.

examples/datadog-express-apm/host-agent/utils/response.js
ajimae: Oh, okay. I guess we should find a way to pass the request method down to this function then.

examples/datadog-express-apm/host-agent/src/sdk-v3.js
ajimae: I don't think it is necessary to have the tracer.js anymore. if we are importing the `dd-trace` then keeping the `tracer.js` around is redundant. Also, this is the option being passed to the `createTelemetryMiddleware` so it doesn't matter, the customer can change it as they wish, they don't need the `tracer.js` file to be able to make these changes.

ajimae: > About can you also update all the require statements to use this lazy import., what do you mean? could you give me an example?

I think you can leave this out, this might require you to change many other unrelated files, as I can now see it's a commonjs file and changing all the `require` to `import` might make this too much unnecessary work. So just disregard it totally.

examples/datadog-express-apm/host-agent/utils/response.js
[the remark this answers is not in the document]
barbara79: I think you are right, I was confused between this custom metric and the general one that already gives back the method. Instead here we need to define it.
I'm gonna find a way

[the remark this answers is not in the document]
ajimae: This should work:

```ts
http_method: res.req.method
```

packages/ts-sdk-apm/src/apm.ts
lojzatran: @barbara79 can you link the branch in the docs here so I know you did it and can review it? 🙏

packages/ts-sdk-apm/src/apm.ts
lojzatran: @barbara79 can you check this comment? 🙇

packages/ts-sdk-apm/src/apm.ts
barbara79: I changed it with this part 

```
console.warn(
        'APM is missing or not a function. Proceeding without tracing.'
      )
      options.apm = () => {}
      
```
      please see the updated code and also the comment made by @ajimae

examples/datadog-express-apm/host-agent/utils/response.js
```diff
@@ -30,7 +30,7 @@ class ResponseHandler {
...
+
```
ajimae, line 33:
```suggestion
http_method: res.req.method,
```

packages/ts-sdk-apm/test/apm.test/apm.test.ts
barbara79: it was failing because we didn't have default values. Let me try to add it again.

examples/datadog-express-apm/host-agent/utils/response.js
barbara79: Oh I didn't see this anyway I did the same liek your suggestion ;)

examples/datadog-express-apm/host-agent/utils/response.js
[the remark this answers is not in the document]
barbara79: Just find out your suggestion :D anyway I made the same change like you ;)

packages/ts-sdk-apm/src/apm.ts
barbara79: I am gonna add the link soon.

barbara79: https://github.com/commercetools/commercetools-docs/blob/update-ts-docs/websites/sdk/src/content/observability/datadog.mdx

lojzatran: I see it's not updated yet?

ajimae (approved):
Great job @barbara79 🚀 
I have approved, however please try and fix the merged conflicts.

Thanks once again for this PR 🥂

PR #924, opened 2025-02-05
[Feat][DEVX-424] Map Error Code to commercetools HTTP Error Code

ajimae:
### Summary
Map error response code field to return commercetools HTTP error response code property

### Completed Tasks
- [x] map code field to http error code
- [x] remove unused code and comments
- [x] minor refactor
- [x] add tests


packages/platform-sdk/test/integration-tests/sdk-v3/middlewares.test.ts
```diff
@@ -115,23 +115,26 @@ describe('Concurrent Modification Middleware', () => {
...
-              action: 'changeName',
-              name: { en: 'test-name' + new Date().getTime() },
-            },
-          ],
-        },
-      })
-      .execute()
-
-    expect(productUpdateResponse.statusCode).toBe(200)
+    try {
```
lojzatran, line 118: Why did you add try/catch? The test should actually fail if there is an exception.

packages/sdk-client-v3/src/types/types.d.ts
```diff
@@ -18,13 +18,13 @@ export type MiddlewareResponse<T = unknown> = {
 
 export type HttpErrorType = {
   name?: string
-  message: string
-  code?: number
+  message?: string
+  code?: string
```
lojzatran, line 22: In the JIRA task this is written: `We can’t change the code field to string as it would be a breaking change.`

lojzatran (commented):
Good job 👍 

I'm lacking a context here. Why are we changing the error structure? Was someone complaining? There is no support issue linked to the JIRA task.

Also it's not clear what is the decision for this task in JIRA. Please add the final decision there if it has been made.

packages/sdk-client-v3/src/types/types.d.ts
ajimae: That was my suggestion, I even went ahead to suggest we add a new field for it, but you specifically suggested that since it wasn't documented, we can go ahead and make the change. Have you forgotten? 🙃

ajimae:
> I'm lacking a context here. Why are we changing the error structure? Was someone complaining? There is no support issue linked to the JIRA task.

Not really changing, there are some duplicate fields I was trying to remove and also, it's not fully following the structure of the original response object. Anyways, I guess we will have to do this in a different task.

> Also it's not clear what is the decision for this task in JIRA. Please add the final decision there if it has been made.
I will definitely update the JIRA ticket to reflect the correct context.

packages/sdk-client-v3/src/middleware/create-http-middleware.ts
```diff
@@ -83,33 +88,30 @@ async function executeRequest({
...
+      e.response?.status ||
+      e.response?.statusCode ||
+      e.response?.data.statusCode ||
+      0
     const message = e.response?.data?.message
 
     const error: HttpErrorType = createError({
       statusCode,
-      code: statusCode,
+      code: 'NetworkError',
```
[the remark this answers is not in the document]
ajimae: That will be an overkill for a single object value, besides if you look at the comment on line 93 you will see why it's obvious it's a `NetworkError`.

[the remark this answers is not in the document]
barbara79: Regarding line 93, while the context is clear here, the same hardcoded string is likely used elsewhere. My concern is the overall error codes definition: having hardcoded strings increase the risk of typos and inconsistency. A centralized constant/enum would be better.

PR #929, opened 2025-02-07
DEVX-533 Newrelic demo app v3

barbara79:


examples/newrelic-express-apm/README.md
```diff
@@ -12,7 +12,7 @@ Example to show how the Newrelic APM can be used in the TypeScript SDK.
 ## Installation
 
 1. Clone/Download the example folder.
-2. Navigate to the path `newrelic-express-apm/host-agent/`.
+2. Navigate to the path `newrelic-express-apm/`.
 3. Create a `.env` file in this path, add and update the content of `.env.sample`, the `DD_API_KEY` is required
```
ajimae, line 16: I think this `DD_API_KEY` should be removed.

examples/newrelic-express-apm/.env.sample
```diff
@@ -1,5 +1,5 @@
 PORT=8000
-NEW_RELIC_APP_NAME=demo-monitoring-app
+NEW_RELIC_APP_NAME=newrelic-express-apm
 NEW_RELIC_LICENSE_KEY=<new_relic_api_key>
```
lojzatran, line 3: Does this app uses newrelic over OTEL or is it without OTEL? I'm asking because you have 2x <new_relic_api_key> and I'm wondering if it could be a duplicate.

examples/newrelic-express-apm/src/sdk-v3/sdk.js
```diff
@@ -31,7 +31,11 @@ const httpMiddlewareOptions = {
 // newrelic options
 const telemetryOptions = {
   createTelemetryMiddleware,
+  userAgent: 'typescript-sdk-middleware-newrelic',
   apm: () => agent,
+  customMetrics: {
```
lojzatran, line 36: This option is not described in the docs: https://docs.commercetools.com/sdk/observability/newrelic. Please add it with explanation what it means.

lojzatran (commented):
Good job updating the demo app 💯 

Looking at the custom metrics in the docs, it is not clear if they are added automatically if nodejs agent is added. I would doubt that since nodejs agent is from Newrelic and the custom metrics are from CT. So this sentence is very confusing: `Include the 
[New Relic Node.js Agent](https://docs.newrelic.com/docs/apm/agents/nodejs-agent/installation-configuration/install-nodejs-agent)
 to add the following custom metrics:`. Please check it and update it if necessary.

https://docs.commercetools.com/sdk/observability/newrelic#add-the-new-relic-agent-configuration

examples/newrelic-express-apm/.env.sample
```diff
@@ -1,5 +1,5 @@
 PORT=8000
-NEW_RELIC_APP_NAME=demo-monitoring-app
+NEW_RELIC_APP_NAME=newrelic-express-apm
```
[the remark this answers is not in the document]
barbara79: it's better to have consistency in all the application and have the same name as the package.json

examples/newrelic-express-apm/.env.sample
barbara79: you are right I didn't notice it. :) 
I need to ask to @ajimae

examples/newrelic-express-apm/src/sdk-v3/sdk.js
barbara79: Yes, it's included in the branch where I am working on ;) 

Please check https://github.com/commercetools/commercetools-docs/tree/update-ts-docs

examples/newrelic-express-apm/README.md
barbara79: I am adding the one for NewRelic

examples/newrelic-express-apm/.env.sample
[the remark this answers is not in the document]
ajimae: I think we should leave this name as it is. If we should make it consistent, then I would suggest we rename others to match this name.

This will require anyone who has run this app to login into Newrelic and rename all their app instances, which is not ideal.

So my suggestion is to leave it as is, consistency here is of very little significance.

examples/newrelic-express-apm/.env.sample
barbara79: @ajimae could you please reply to Lam question?

examples/newrelic-express-apm/.env.sample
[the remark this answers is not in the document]
ajimae: Exactly why not renaming it won't be a big of a deal, what do you think? 🙃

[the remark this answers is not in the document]
barbara79: You are not replying to my suggestion with a real reason 🙃 
it's to have not a generic name but something that identify the monitoring tool used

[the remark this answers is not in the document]
lojzatran: Yes, the name for datadog looks like generated from a nodejs itself. The name was `my-app` and it was in package.json, so I suggested to rename it: https://github.com/commercetools/commercetools-sdk-typescript/pull/919/files/6b69538216a2cf6e377ed469e83952b4fee8fd04#diff-f8ae131f74cfff8aa9a1d6d54b09f68c3563591668e2afdd77ea224fbd20e90b

But in general it's not so critical. If renaming does more harm, then you can leave the original name.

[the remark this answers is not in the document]
ajimae: > You are not replying to my suggestion with a real reason 🙃 it's to have not a generic name but something that identify the monitoring tool used

I actually gave a reason here:

> This will require anyone who has run this app to login into Newrelic and rename all their app instances, which is not ideal.

If for example I want to run this app, I will have to log into my newrelic account and rename application name to the new name, which is why I said it is not ideal. 🙂

examples/newrelic-express-apm/.env.sample
ajimae: @lojzatran you can optionally use it with OTEL. You only have to pass it as the command line argument stated [here](https://docs.commercetools.com/sdk/observability/newrelic#add-the-new-relic-agent-configuration). I guess maybe we need more documentation on this one.

examples/newrelic-express-apm/.env.sample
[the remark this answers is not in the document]
barbara79: I understand your concern but I think we are putting our focus on the wrong thing.
I am going to rollback my change to proceed with the PR   because it contains valuable changes, however we need to address our naming consistency sooner rather than later. Thanks for your feedback.

examples/newrelic-express-apm/.env.sample
barbara79: yes, we do it's not clear that we could do it with OTEL as well.
Since I have an opened branch for the official documentation I am gonna include it

barbara79: I don't find any span related to OTEL here, could you help me please?
Not sure if it's working, then I need to add something in the official documentation

barbara79: I'm gonna open a separated task for it and the related documentation

lojzatran (approved):

PR #968, opened 2025-03-10
[Chore] Rename Colliding Package Names

ajimae:
### Summary
Rename packages with the same (colliding) names

### Completed Task
- [x] rename packages with colliding names

### Screenshot

Screenshot 2025-03-10 at 10 45 42

### DIFF
--- a/examples/datadog-express-apm/container-agent/package.json
+++ b/examples/datadog-express-apm/container-agent/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "datadog-express-apm",
+  "name": "container-agent",
   "version": "1.0.0",
   "main": "server.js",
   "license": "MIT",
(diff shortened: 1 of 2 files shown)


examples/datadog-express-apm/container-agent/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "datadog-express-apm",
+  "name": "container-agent",
```
barbara79, line 2: I believe it's better if you add something related to Datadog like "datadog-container-apm"

examples/datadog-express-apm/host-agent/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "datadog-express-apm",
+  "name": "host-agent",
```
barbara79, line 2: I believe it's better if you add something related to Datadog like "datadog-host-apm"

ajimae: It is already in a directory called `examples/{datadog-express-apm}/host-agent/package.json` so it's irrelevant here.

PR #978, opened 2025-03-12
DEVX-534 Migrate  Dynatrace demo app to client v3

barbara79:


examples/dynatrace-express-apm/src/sdk-v3/sdk.js
```diff
@@ -1,46 +1,44 @@
...
 const fetch = require('node-fetch')
+const path = require('path')
 
 const oneagent = path.join(__dirname, '..', '..', 'oneagent-tracer.js')
+const tracer = require(oneagent)
+
 const projectKey = process.env.CTP_PROJECT_KEY
 const authMiddlewareOptions = {
-  host: 'https://auth.europe-west1.gcp.commercetools.com',
+  host: process.env.CTP_AUTH_URL,
```
ajimae, line 12: Please update the `.env.example` file to include this new variable

examples/dynatrace-express-apm/src/sdk-v3/sdk.js
```diff
@@ -1,46 +1,44 @@
...
-    },
   },
   scopes: [`manage_project:${projectKey}`],
-  fetch,
+  httpClient: fetch,
 }
 
 const httpMiddlewareOptions = {
-  host: 'https://api.europe-west1.gcp.commercetools.com',
+  host: process.env.CTP_API_URL,
```
ajimae, line 23: Please update the `.env.example` file to include this new variable

examples/dynatrace-express-apm/src/sdk-v3/sdk.js
```diff
@@ -1,46 +1,44 @@
...
-  // apm: () => {}
+  customMetrics: {
+    newrelic: false,
+    datadog: false,
+  },
 }
 
 const client = new ClientBuilder()
-  .withPasswordFlow(authMiddlewareOptions)
+  .withClientCredentialsFlow(authMiddlewareOptions)
```
ajimae, line 41: Why are we changing this to a client credentials flow?

examples/dynatrace-express-apm/README.md
```diff
@@ -0,0 +1,30 @@
...
+- A Composable Commerce Project with a configured [API Client](https://docs.commercetools.com/sdk/js-sdk-getting-started).
+- Your Project must have existing Products containing Variants, and at least one Customer.
+- If your Project is currently empty, you can install the [SUNRISE sample data](https://github.com/commercetools/commercetools-sunrise-data).
+- For Dynatrace setup, follow the instructions stated in our [official documentation website](https://docs.commercetools.com/sdk/observability/dynatrace#typescript-sdk) to properly install and set up the Dynatrace agent.
+
+## Installation
+
+1. Clone/Download the example folder.
+2. Navigate to the path `dynatrace-express-apm`.
+3. Create a `.env` file in the src folder by copying the contents of `.env.sample`, then update the values as needed. The variables required to run the demo app are: `CTP_PROJECT_KEY`, `CTP_CLIENT_ID`, `CTP_CLIENT_SECRET`,`CTP_AUTH_URL`,`CTP_API_URL` and `DYNATRACE_ENV_ID` (Dynatrace environment ID) and `DT_OPEN_TOKEN` (Dynatrace Token).
```
lojzatran, line 16:
```suggestion
3. Create a `.env` file in the src folder by copying the contents of `.env.sample`, then update the values as needed. The variables required to run the demo app are: `CTP_PROJECT_KEY`, `CTP_CLIENT_ID`, `CTP_CLIENT_SECRET`,`CTP_AUTH_URL`,`CTP_API_URL` and `DT_ENV_ID` (Dynatrace environment ID) and `DT_OPEN_TOKEN` (Dynatrace Token).
```

examples/dynatrace-express-apm/src/sdk-v3/sdk.js
jenschude: I would say for simplicity when no login is needed

barbara79: yes, that was the reason :)

examples/dynatrace-express-apm/README.md
barbara79: yes, I need to change it in the other files, but yes, it's the same variable name that I found in Dynatrace documentation.
Just to avoid confusion I am gonna update it.
Thanks

examples/dynatrace-express-apm/src/sdk-v3/sdk.js
barbara79: you are right, I forgot to update it

examples/dynatrace-express-apm/src/sdk-v3/sdk.js
barbara79: you are right, I forgot to update it

examples/dynatrace-express-apm/src/sdk-v3/sdk.js
ajimae: We have an [endpoint](https://github.com/commercetools/commercetools-sdk-typescript/blob/62fb08d43dbcd7713cf8e49864a511a15a8b2299/examples/dynatrace-express-apm/src/service/CartService.js#L28) that required this, that is why we used the password flow originally, except there is a way to do this without the password flow, else this endpoint will fail.

I will still maintain we revert this change back to use the password flow.

examples/dynatrace-express-apm/src/sdk-v3/sdk.js
ajimae: We have an [endpoint](https://github.com/commercetools/commercetools-sdk-typescript/blob/62fb08d43dbcd7713cf8e49864a511a15a8b2299/examples/dynatrace-express-apm/src/service/CartService.js#L28) that required this, that is why we used the password flow originally, except there is a way to do this without the password flow, else this endpoint will fail.

I will still maintain we revert this change back to use the password flow.

barbara79:
> Is it possible to add a screenshots of the records on the Dynatrace dashboard?

You can see the results on the internal [wiki page created ](https://commercetools.atlassian.net/wiki/spaces/CTCLIENTS/pages/1752301612/Dynatrace+Metrics+for+TS+SDK)

examples/datadog-express-apm/host-agent/env.sample
```diff
@@ -4,7 +4,8 @@ CTP_CLIENT_ID=
 CTP_CLIENT_SECRET=
 CTP_AUTH_URL=https://auth.<your-region>.gcp.commercetools.com
 CTP_API_URL=https://api.<your-region>.gcp.commercetools.com
-
+CTP_CLIENT_USERNAME=<customer_username>
+CTP_CLIENT_PASSWORD=<customer_password>
```
ajimae, line 8: I believe these environment entries are not needed here. Can we remove them to avoid confusion?

ajimae (approved):
LGTM

Issue #1006, opened 2025-03-28
process.version is not available at React Native runtime


professorkolik:
**Describe the bug**

Initialize ClientBuilder on ReactNative app throws an error  `[TypeError: Cannot read property 'slice' of undefined]`

It's thrown from `packages/sdk-client-v3/src/utils/userAgent.ts`

```
const nodeVersion: string = process?.version.slice(1) || 'unknow' // unknow environment like  React Native etc
```
process variable is not available on React Native runtime.

**Steps to reproduce**

- Dependencies:
```
"@commercetools/platform-sdk": "^8.5.0",
"@commercetools/ts-client": "^3.2.0",
"react-native": "0.75.4",
```

- buildClient.ts
```
import type {
  HttpMiddlewareOptions,
  AuthMiddlewareOptions,
} from '@commercetools/ts-client';
import { ClientBuilder } from '@commercetools/ts-client';
import Config from 'react-native-config';

const projectKey = Config.CTP_PROJECT_KEY;
const scopes = Config.CTP_SCOPES.split(' ');

// Configure authMiddlewareOptions
const authMiddlewareOptions: AuthMiddlewareOptions = {
  host: Config.CTP_AUTH_URL,
  projectKey,
  credentials: {
    clientId: Config.CTP_CLIENT_ID,
    clientSecret: Config.CTP_CLIENT_SECRET,
  },
  scopes,
  httpClient: fetch,
};

// Configure httpMiddlewareOptions
const httpMiddlewareOptions: HttpMiddlewareOptions = {
  host: Config.CTP_API_URL,
  httpClient: fetch,
};
export const ctpClient = new ClientBuilder()
  .withClientCredentialsFlow(authMiddlewareOptions)
  .withHttpMiddleware(httpMiddlewareOptions)
  .withConcurrentModificationMiddleware()
  .build();
```

**Expected behavior**

SDK does not throw an error on ReactNative

**How to solve quick win**

const nodeVersion: string = process?.version?.slice(1) || 'unknow' // unknow environment like  React Native etc

**How to solve in proper way**

Give ability to consumer to define "systemInfo"

ajimae:
Hi @professorkolik 

This should now be fixed, please update to this [latest](https://www.npmjs.com/package/@commercetools/ts-client/v/3.2.1) version and let us know if it works for you.


Thank you

professorkolik:
Thank you @ajimae it works

PR #1009, opened 2025-03-28
[Fix] Issue with `process` Object in a non supported environments e.g React Native

ajimae:
### Summary
Fix issue with missing `process` object in non supported environments like `React Native`

### Completed Tasks
- [x] fix issue process object in an unknown enviornment
- [x] assert that the process object exists and contains the necessary properties

### Related Issue
[1006](https://github.com/commercetools/commercetools-sdk-typescript/issues/1006)

### DIFF
--- a/packages/sdk-client-v3/src/utils/userAgent.ts
+++ b/packages/sdk-client-v3/src/utils/userAgent.ts
@@ -16,8 +16,8 @@ const isBrowser = (): boolean =>
 function getSystemInfo(): string {
   if (isBrowser()) return window.navigator.userAgent
 
-  const nodeVersion: string = process?.version.slice(1) || 'unknow' // unknow environment like  React Native etc
-  const platformInfo = `(${process.platform}; ${process.arch})`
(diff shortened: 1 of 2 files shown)


packages/sdk-client-v3/src/utils/userAgent.ts
```diff
@@ -16,10 +16,10 @@ const isBrowser = (): boolean =>
 function getSystemInfo(): string {
   if (isBrowser()) return window.navigator.userAgent
 
-  const nodeVersion: string = process?.version.slice(1) || 'unknow' // unknow environment like  React Native etc
-  const platformInfo = `(${process.platform}; ${process.arch})`
+  const nodeVersion: string = process?.version?.slice(1) || 'unknow' // unknow environment like  React Native etc
```
professorkolik, line 19: @ajimae Hey, thanks a lot for quick fix

Would you be able to fix typo as well please, on next planned release?

`unknow` => `unknown`

PR #1130, opened 2025-07-10
[Chore][DEVX-571]: Move CI from CircleCI to Github Actions

ajimae:
### Summary
Migrate all related CI/CD workflows from CircleCI to Github Actions

### Completed Tasks
- [x] move all related CI-CD workflows from circleci to github actions


.github/actions/ci/action.yml
```diff
@@ -0,0 +1,26 @@
+name: CI
+
+description: Shared action to install dependencies
+
+runs:
+  using: composite
+
+  steps:
+    - name: Read .nvmrc
```
jenschude, line 9: would it make sense to add the checkout step here also, as it would ensure that the .nvmrc is present

jenschude (approved):
👍 
especially the action for setting up node etc

.github/actions/ci/action.yml
ajimae: Not really, it is not necessary here as the repo is already checked-out in the step before. Also we are only just reading out the `.nvmrc` and making the version within available in the environment.

PR #1154, opened 2025-08-21
[Fix] Add Meta Headers to Auth Requests

ajimae:
### Summary
Add meta headers to all auth http requests

### Completed Tasks
- [x] include meta-headers to as part of auth http requests in `authProcessor` function


packages/sdk-client-v3/src/middleware/auth-middleware/auth-request-executor.ts
```diff
@@ -64,6 +66,7 @@ export async function executeRequest(options: ExecuteRequestOptions) {
       url,
       method: 'POST',
       headers: {
+        ...request.headers,
```
qmateub, line 69: This totally makes sense and explains why the headers were not present on the auth calls! 👏

Issue #1273, opened 2026-05-11
UUID lib vulnerability


drazenhrgar:
Hi!

Our security scanning tools are reporting critical vulnerability with UUID lib included in the `@commercetools/ts-sdk-apm`

Category CWE-787 | Out-of-bounds Write

CVE Details
Published: Apr 24, 2026
uuid is for the creation of RFC9562 (formerly RFC4122) UUIDs. 11.0.0 prior to 11.1.1, 12.0.0 prior to 12.0.1, 13.0.0 prior to 13.0.1, v3, v5, and v6 accept external output buffers but do not reject out-of-range writes (small buf or large offset). This allows silent partial writes into caller-provided buffers. This vulnerability is fixed in 11.1.1, 12.0.1, 13.0.1, 14.0.0.

ajimae:
Thanks for reporting, we will take a look right away.

ajimae:
This is now done, a new version has been release and can be found [here](https://www.npmjs.com/package/@commercetools/ts-sdk-apm/v/4.4.2)

ajimae:
I will go ahead and close this issue.

PR #1274, opened 2026-05-11
[Chore] Remove UUID npm package

ajimae:
### Summary
Remove `UUID` npm package entirely

### Completed Tasks
- [x] remove the uuid package
- [x] add a helper to generate a uuid-like string

### Related Issue
#1273

### DIFF
--- a/packages/ts-sdk-apm/package.json
+++ b/packages/ts-sdk-apm/package.json
@@ -13,8 +13,7 @@
     "@opentelemetry/exporter-metrics-otlp-http": "^0.205.0",
     "@opentelemetry/sdk-node": "^0.205.0",
     "dd-trace": "^5.31.0",
-    "newrelic": "^13.0.0",
-    "uuid": "13.0.0"
+    "newrelic": "^13.0.0"
   },
   "publishConfig": {
     "access": "public"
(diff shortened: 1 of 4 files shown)


packages/ts-sdk-apm/package.json
```diff
@@ -13,8 +13,7 @@
     "@opentelemetry/exporter-metrics-otlp-http": "^0.205.0",
     "@opentelemetry/sdk-node": "^0.205.0",
     "dd-trace": "^5.31.0",
-    "newrelic": "^13.0.0",
-    "uuid": "13.0.0"
```
drazenhrgar, line 17: why not just bumping the version? 

also, in our lib, the UUID that is referenced in `ts-sdk-apm` is `11.1.0` and we are using `@commercetools/ts-sdk-apm v4.4.1`

maybe the issue just that you don't have the latest code in the release package?

ajimae: We are now completely dropping all reference to the `UUID` package. In the main `SDK` we did away with the package as well.

Does your application mandatorily need to have this `UUID` package?

drazenhrgar: no it's fine, we can include it manually if we need it, was just checking why would you rewrite a logic. Thx

PR #1312, opened 2026-06-02
[Chore] Update dependencies

ajimae:
### Summary
Update and upgrade dependencies

### Completed Tasks
- [x] update and upgrade dependencies
- [x] ignore the examples in renovate and dependabot configs


packages/checkout-sdk/src/generated/models/application.ts
```diff
@@ -467,8 +466,7 @@ export interface AddCountryUpdateAction extends IApplicationUpdateAction {
    */
   readonly country: string
 }
-export interface RemoveAllowedOriginUpdateAction
-  extends IApplicationUpdateAction {
+export interface RemoveAllowedOriginUpdateAction extends IApplicationUpdateAction {
```
ShipilA, line 469: why did you change it?

ajimae: It didn't really change, it is just the code formatting that changed so it's now inline instead of multi-line

geethanga-ct (commented):
@ajimae in the generated code I can see a lot of formatting changes which has not impact on the eventual outcome of the code. Do you know why these formatting changes appeared? Was it because of the prettier version update?

ajimae:
> @ajimae in the generated code I can see a lot of formatting changes which has not impact on the eventual outcome of the code. Do you know why these formatting changes appeared? Was it because of the prettier version update?

Have to do with some configuration changes due to packages and dependencies update as well.

ShipilA (approved):

PR #1442, opened 2026-08-07
[Feat][DEVX-851] Make Platform SDK Use Same Client Host URL

ajimae:
### Summary
Make the platform sdk use same host url as that passed into the `httpMiddlewareOptions`

### Completed Tasks
- [x] pass the host url to the platform sdk from the http client configuration


geethanga-ct (approved):
LGTM.

PR #1444, opened 2026-08-10
[Feat][DEVX-850] Add @commercetools/graphql-sdk

ajimae:
## Summary

Adds `@commercetools/graphql-sdk`: GraphQL support for the TypeScript SDK. One package, one entry point, one client.

The generated SDKs already expose the GraphQL endpoint through `apiRoot.graphql().post(...)`, but that call takes a `query` string and hands back `data: any`. This package keeps the same transport — the same client, authentication, middlewares, retries and correlation ids — and adds the typing on top.

## Usage

```ts
import { createApiBuilderFromCtpClient } from '@commercetools/platform-sdk'
import { createGraphQLClient } from '@commercetools/graphql-sdk'

const apiRoot = createApiBuilderFromCtpClient(client).withProjectKey({ projectKey })
const graphQL = createGraphQLClient(apiRoot)

const data = await graphQL
  .query({
    customers: (customers) =>
      customers
        .where('firstName="Martha"')
        .sort(['lastName asc'])
        .limit(20)
        .total()
        .results((result) => result.firstName().lastName().email()),
  })
  .executeOrThrow()

data.customers.total            // number
data.customers.results[0].email // string
```

Field names, arguments and the shape of the result all come from the committed GraphQL schema, so no GraphQL string is written by hand and a typo is a compile error. Autocomplete works at every step of the chain.

Arguments become GraphQL variables, never string interpolation:

```graphql
query ($v1: String, $v2: [String!], $v3: Int) {
  customers(where: $v1, sort: $v2, limit: $v3) {
    total
    results { firstName lastName email }
  }
}
```

## The API surface

| Method                          | Use                                                                                 |
| ------------------------------- | ----------------------------------------------------------------------------------- |
| `createGraphQLClient(apiRoot)`  | creates the client on top of an existing api root, or a `graphql()` request builder   |
| `query(selectors)`              | builds a query from the schema                                                       |
| `mutate(selectors)`             | builds a mutation, same chaining syntax                                              |
| `rawQuery(document, options)`   | escape hatch for an operation the builder does not express                           |

`rawQuery()` takes a query string or any `TypedDocumentNode` — the interface produced by GraphQL Code Generator and gql.tada — and returns the same `GraphQLApiRequest` through the same transport, so an existing codegen setup keeps its types.

## Design notes

**Only the fields you project are in the result type.** Reading a field that was never selected is a compile error, not `undefined` at runtime.

**Arguments and fields chain the same way**, and which one a call is comes from the schema:

```ts
products
  .limit(5)                          // an argument of `products`
  .total()                           // a field of ProductQueryResult
  .results((product) => product.id())
```

**A localized field takes its argument on its own chain**, because the locale belongs to the field and not to the query around it:

```ts
current.name((name) => name.locale('en'))          // -> name(locale: $v1)
current.name()                                     // -> name
current.nameAllLocales((n) => n.locale().value())  // -> nameAllLocales { locale value }
```

**Nothing is sent until `execute()`.** `query()`, `mutate()` and `rawQuery()` return a `GraphQLApiRequest` mirroring the generated `ApiRequest`, with `clientRequest()` and `requestBody()` for logging and tests.

**Errors.** The endpoint answers HTTP 200 for failed operations, so `execute()` resolves with the full response and never rejects for them. `executeOrThrow()` resolves with `body.data` and rejects with a `GraphQLRequestError` carrying `errors`, any partial `data`, and the full `response`.

## Layout

```
packages/graphql-sdk/
  schema.graphqls              committed GraphQL schema, the codegen input
  src/index.ts                 the only entry point
  src/client/graphql-client.ts GraphQLClient: query / mutate / rawQuery
  src/client/graphql-api-request.ts
  src/builder/chain.ts         the fluent projection builder
  src/builder/generated/       genql output, regenerated from schema.graphqls
  src/errors/ src/types/ src/utils/
```

Regenerating after a schema update:

```bash
make gen_graphql_builder
```

This runs [genql](https://genql.dev) over `schema.graphqls` and rewrites `src/builder/generated`. Scalars are mapped to their natural TypeScript types (`Long` → `number`, `Country`/`Locale` → `string`, `Json` → `unknown`, and so on).

Not wired into `make build`: that chain is driven by the RAML specs, while `schema.graphqls` tracks the GraphQL API and is refreshed independently.

## Verification

- `yarn build` clean, `tsc --noEmit` clean on the package
- 44 package tests pass; full suite 459/460 (the one failure, `sdk-client-v3` `retryOnAbort`, is a pre-existing network flake and reproduces on a clean tree)
- Nine representative documents built by the builder — products, project, customers, carts, orders, categories, product search, a multi-root query and a mutation — were validated against the official commercetools GraphQL schema and all came back valid
- Structural compatibility with `@commercetools/platform-sdk` and `@commercetools/history-sdk` is pinned against the real generated code rather than a hand written mock (`tests/generated-sdk-compatibility.test.ts`)

### Ticket
[DEVX-850](https://commercetools.atlassian.net/browse/DEVX-850)

[DEVX-850]: https://commercetools.atlassian.net/browse/DEVX-850?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jenschude (approved):
I recommend to a prerelease with version 0.1.0 for this package. In order to better test it out and see if some functionality may be missing

ajimae:
> I recommend to a prerelease with version 0.1.0 for this package. In order to better test it out and see if some functionality may be missing

Exactly, I have a [beta](https://www.npmjs.com/package/@commercetools/graphql-sdk) release on NPM already and this first version is a minor release, see the changeset section above.

|Name	|Type|
|-----|----|
|@commercetools/graphql-sdk |Minor|

Once we certify that it is all good, we can then do the first major (v1.0.0) release.

ShipilA (approved):
Looks great to me

PR #1452, opened 2026-08-16
[Feat][DEVX-856] Add a Fluent Builder For The `where` Predicate

ajimae:
## Summary

`where` on a paged query field keeps taking a [Query Predicate](https://docs.commercetools.com/api/predicates/query) string exactly as before — that stays the escape hatch for anything the builder does not express — and now also takes a callback that builds one through a checked builder:

```ts
const data = await graphQL
  .query({
    customers: (customers) =>
      customers
        .where((customer) =>
          customer.firstName
            .is('Martha')
            .and(customer.addresses((address) => address.city.is('Berlin')))
            .and(customer.customerGroup((group) => group.id.is(goldGroupId)))
        )
        .sort(['lastName asc'])
        .limit(20)
        .results((result) => result.firstName().lastName().email()),
  })
  .executeOrThrow()
```

which sends

```
(firstName = "Martha") and (addresses(city = "Berlin")) and (customerGroup(id = "..."))
```

Field names come from the resource being filtered, and the operators from what the field holds, so a predicate the API would reject is a compile error. That includes the two that are easy to get wrong: a Reference exposes only `id`/`typeId` inside a predicate, never the fields of the resource it points at, and a localized field is addressed per locale.

The builder is derived from the response types genql already generates, so there is no second artefact to keep in step with the schema.

## Changes

- `src/builder/where.ts` (new) — the predicate types and the runtime proxy that emits the string.
- `src/builder/chain.ts` — threads a `TResult` type parameter through the chain so `where` knows which resource it filters over, and resolves a `where` callback at runtime. Defaults to `unknown`, so every other member behaves exactly as before.
- `src/client/graphql-client.ts` — pairs the root selection type with `Query`/`Mutation` so that `TResult` has something to start from.
- `src/index.ts` — exports `not` and the predicate types.
- README section and a changeset.
- Drops the comparisons with the Java and .NET SDKs from the README and the doc comments, so the documentation stands on its own.

## Known limits

Both are documented in the README:

- A handful of schema fields have no counterpart in the response representation a predicate is evaluated against, and are offered even though the API will reject them.
- Query endpoints restrict predicates to a documented subset of fields, which is not machine readable, so the builder does not enforce it.

## Testing

- `packages/graphql-sdk/tests/where.test.ts` (new) covering the emitted strings for each operator, nesting, references, localized fields, Custom Fields, escaping and grouping.
- `yarn jest packages/graphql-sdk` — 88 passed, 5 suites.
- `yarn typecheck` — clean.

## Ticket
[DEVX-856](https://commercetools.atlassian.net/browse/DEVX-856)

[DEVX-856]: https://commercetools.atlassian.net/browse/DEVX-856?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/plenty-lions-argue.md
+++ b/.changeset/plenty-lions-argue.md
@@ -16,4 +16,6 @@ customers.where((customer) =>
 ```
 
 Field names come from the resource, operators from what the field holds, so a predicate the API
-would reject is a compile error.
+would reject is a compile error. Attribute and Custom Field values are descended into for the
+types the schema cannot express (`value(centAmount = 999 and currencyCode = "EUR")`), and a
+GeoJSON field is matched with `withinCircle(longitude, latitude, radiusInMeters)`.
--- a/.changeset/plenty-lions-argue.md
+++ b/.changeset/plenty-lions-argue.md
@@ -1,0 +1,19 @@
+---
+'@commercetools/graphql-sdk': patch
+---
+
+Add an optional fluent builder for the `where` query predicate.
+
+`where` keeps taking a predicate string exactly as before, and now also takes a callback that
+builds one through a checked builder derived from the resource being filtered:
+
+```ts
+customers.where((customer) =>
+  customer.firstName
+    .is('Martha')
+    .and(customer.addresses((address) => address.city.is('Berlin')))
+)
+```
+
+Field names come from the resource, operators from what the field holds, so a predicate the API
+would reject is a compile error.
--- a/packages/graphql-sdk/README.md
+++ b/packages/graphql-sdk/README.md
@@ -1,14 +1,12 @@
 # @commercetools/graphql-sdk
 
-GraphQL support for the commercetools TypeScript SDK, with a schema derived query builder.
-
-This is the counterpart of the Java SDK's `commercetools-graphql-api` module and the .NET SDK's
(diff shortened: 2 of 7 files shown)


kellydavid (commented):
This is looking very nice!

I did a review (with claude) and found a couple of issues:

 - Product attributes are exposed as JSON in the GQL API and so if the value has some nested fields (eg. a Money value) then the builder doesn't support matching those nested fields.
 - geoLocation filtering not included in the predicate builder

I would suggest adding a test case for both of these scenarios above.

ajimae:
> This is looking very nice!
> 
> I did a review (with claude) and found a couple of issues:
> 
> * Product attributes are exposed as JSON in the GQL API and so if the value has some nested fields (eg. a Money value) then the builder doesn't support matching those nested fields.
> * geoLocation filtering not included in the predicate builder
> 
> I would suggest adding a test case for both of these scenarios above.

Addressed in this commit [08789d5](https://github.com/commercetools/commercetools-sdk-typescript/pull/1452/commits/08789d5ca8d8d21af5a23d446560e19620934313)

kellydavid (approved):

jenschude (approved):
That's cool 👍

